PR #152, opened 2012-03-09
Full Proxy support via environment variables

shogo82148:
Hello
I wrote proxy support and fixed authentication error for streaming.
- improved error handling for APIMethods
- proxy support for streaming
- fixed authentication error for streaming
  - I tryed to connect user stream using Python 2.6.6, but Twitter always returns 401 error.

This is including pull request #80.


pike007:
@joshthecoder  Please share sample code for using tweepy under proxy

PGryllos:
that would be really helpful

PR #282, opened 2013-04-17
Rate limit support, multiple authentication, search using v1.1 api and more.

nirg:
Added multiple features (sorry for not separating commits):
- API now support switching between multiple authentication handler
- Allow blocking when reaching rate limit.
- Search is now using Twitter API v1.1
- 'Pagination' of timelines is now done properly with max_id parameter, applies to all endpoints supporting 'max_id'. Other endpoint that have 'page' in their argument list still use the old pagination iterator.
- Fixed an issue with PageIterator that caused the iterator to make an extra call to the Twitter API beyond the provided page limit.
- Add unit tests for all of the above.

### DIFF
--- a/tweepy/cursor.py
+++ b/tweepy/cursor.py
@@ -150,7 +150,7 @@ class MaxIdIterator(BaseIterator):
         if len(items) < 1: 
             raise StopIteration
         
-        self.kargs['max_id'] = min([t['id'] if 'id' in t else t.id for t in items])-1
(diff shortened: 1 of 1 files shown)


saitam1:
Hi @nirg do you have an idea what I am doing wrong? I am getting an error when trying to switch to the next auth: http://stackoverflow.com/questions/42059191/tweepy-multiple-auth-handler

Issue #291, opened 2013-05-21
406 error in Streaming API when filtering on language


magsol:
I invoke the streaming API:

``` python
stream.filter(languages = ["en"])
```

I get 406 errors. However, if I simply emulate the OAuth streaming example:

``` python
stream.filter(track = ["basketball"])
```

it works fine. I noticed that even though Twitter supposedly [implemented language support](https://dev.twitter.com/calendar) on March 26, it hasn't shown up on their official documentation for the [filter streaming API](https://dev.twitter.com/docs/api/1.1/post/statuses/filter), so I can't tell which side the problem is on. Any suggestions?

It should be noted: I'm using the most current version of tweepy from github, specifically so I can use the language filter.

saifmahamood:
Hi, I'm still supposedly getting 406 errors after I do stream.filter(languages = ["en"]). 

Any help?

FredZHENG:
I guess you need to pass on the "track" parameter as well. 

twitterStream.filter(track = ["car"], languages=["en"]) works for me,

but twitterStream.filter(languages=["en"]) does not.

PR #355, opened 2013-09-08
Added optional language filter on sample endpoint

mike-anderson:
Resolves Issue #291 

I found that filtering on language alone can only be done through the sample endpoint on the streaming API. This problem was brought up in issue 291 (filtering on language alone will result in 406 error), originally caused by twitter not supporting language, but now caused by filter apparently only filtering on track or location.

specifying language in sample seems to fix the problem for me.


saifmahamood:
Hey guys. 

I noticed that in the filter endpoint there exists now an option for languages as well. I tried languages = ["en"] but I still get the 406 error. I appreciate any help.

PR #484, opened 2014-08-25
RateLimitHandler

ducu:
- **Multiple access tokens with `RateLimitHandler`** (https://github.com/svven/tweepy/blob/master/tweepy/limit.py)

> RateLimitHandler class inherits from OAuthHandler, and introduces add_access_token method that can be used as follows:
> 
> ``` python
> from tweepy import RateLimitHandler
> from config import CONSUMER_KEY, CONSUMER_SECRET, ACCESS_TOKENS
> 
> def get_api():
>   auth = RateLimitHandler(CONSUMER_KEY, CONSUMER_SECRET)
>   for key, secret in ACCESS_TOKENS:
>       try:
>           auth.add_access_token(key, secret)
>       except Exception, e:
>           print key, e
>   print 'Token pool size: %d' % len(auth.tokens)
>   api = API(auth, 
>       wait_on_rate_limit=True, wait_on_rate_limit_notify=True)
>   return api
> 
> api = get_api()
> ```
> 
> Provided access tokens are used selectively based on requested resource (https://dev.twitter.com/docs/rate-limiting/1.1/limits) and current rate limits. The access token with most remaining requests per window for the specified resource is being selected and used when applying the authentication, before the actual request is performed. This pattern ensures the usage of available access tokens in a round robin fashion, exploiting to maximum the rate limits.
> 
> Here's a good use case for it: https://github.com/ducu/twitter-most-followed#requirements
- **Various fixes**: IdIterator, response status code, API DELETE methods.

### DIFF
--- a/tweepy/binder.py
+++ b/tweepy/binder.py
@@ -150,15 +150,20 @@ def bind_api(**config):
                     self._reset_time = reset
                     self._remaining_calls = remaining
                 # handle running out of api calls
-                if self.wait_on_rate_limit:
-                    if self._reset_time is not None:
-                        if self._remaining_calls is not None:
-                            if self._remaining_calls == 0:
-                                sleep_time = self._reset_time - int(time.time())
-                                if sleep_time > 0:
-                                    if self.wait_on_rate_limit_notify:
-                                        print "Rate limit reached. Sleeping for: " + str(sleep_time)
-                                    time.sleep(sleep_time + 5)  # sleep for few extra sec
+                if self._reset_time is not None and \
+                    self._remaining_calls is not None and self._remaining_calls == 0:
+                    sleep_time = self._reset_time - int(time.time())
+                    if self.wait_on_rate_limit:
+                        if sleep_time > 0:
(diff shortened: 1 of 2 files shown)


saitam1:
I have an app running, which collects tweets of certain hashtags/keywords and then analyzes them and puts them in a chart. Unfortunately there was a crash 2 days ago. So I need to fill the data gap now manually, but the problem is, that I can't get these tweets, because of the search limits. I found [this stackoverflow question](http://stackoverflow.com/questions/38041037/tweepy-use-multiple-api-keys-with-cursor-to-search-twitter), but I guess the person is using a forked Tweepy version. Do you have some tips on how to solve the problem? @ducu Are your solutions working with Python3 ? I also thought about using [GetOldTweets-python](https://github.com/Jefferson-Henrique/GetOldTweets-python)

ducu:
Hey, nope I didn't check if it's working with Python 3, and the PR was not accepted, but you can try my fork: https://github.com/svven/tweepy

PR #537, opened 2015-01-08
Session errors be gone

wdahlenburg:
Install error as shown below is fixed

Traceback (most recent call last):
  File "setup.py", line 17, in <module>
    install_reqs = parse_requirements('requirements.txt', session=uuid.uuid1())
TypeError: parse_requirements() got an unexpected keyword argument 'session'

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #from distutils.core import setup
-import re, uuid
+import re
 from setuptools import setup, find_packages
 from pip.req import parse_requirements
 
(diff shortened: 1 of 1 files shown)


volnt:
You get this error because you have an old pip version.

To update your pip version you can do `pip install -U pip`.

wordstospend:
You may have found your self in this position if you are
- on a Mac
- Using easy_install to install virtualenv

This results in using virtualenv default version of pip which I believe is at 1.3.1 :(
You can solve this by instead:
- using brew to install the latest 2.7 python ~ 2.7.9 as of this note.
- Brew will also install pip.
- use pip to install virtualenv : pip install virtualenv

You will now be using the latest version of pip, but the global one rather than the virtualenv local one.
NOTE: This is a change from the older easy_install workflow.
NOTE: The above solution works for the same reason. `pip install -U pip` will result in a global pip of updated version.

aspdave:
First of all Thanks a bunch , error is gone after upgrade of pip version. I had this issue in aws ec2 
instance possibly AMI had old version original error was  "ImportError: No module named tweepy" . Re install of tweepy failed with session error. finally everything run fine with advise . thanks again.

PR #545, opened 2015-01-20
fix setup: replace uuid by PipSession

Laisky:
use PipSession rather then uuid


joshthecoder:
What's the reason for this change?

Laisky:
@joshthecoder I think PipSession is much better than uuid. 

[https://github.com/pypa/pip/blob/develop/pip/download.py#L398](https://github.com/pypa/pip/blob/develop/pip/download.py#L398)

PR #547, opened 2015-01-25
Fixes error message on file too big to reflect actual file size limit.

skoczen:
Super simple PR, but something I ran into and was confused by!

Thanks!


joshthecoder:
Good catch. Thanks for fixing it.

PR #548, opened 2015-01-25
More consistent SVG badges

dfm:
This patch uses [shields.io](http://shields.io/) to provide consistently styled vector README badges. Personally, I like this a bit better than the current png badges but that's just me :smile: 

Either way, thanks for this code!


joshthecoder:
Looks good. Thanks!

PR #558, opened 2015-02-16
Fix bug where streams freeze up on Python 3 due to string formatting bug...

### LINKED ISSUE #556: tweepy Stream does not return any data
I am using the example script streaming.py from this repository. When I run the script no data is returned at all, no error code, nothing.

Usingthe REST API with the same credentials I did not experience any problem. All the things I tested returned a result. It's just not working with the streaming api.

I am using Python 3.4 in a virtualenv. I tested with the version from PyPi as well as from the Github Repository by cloning it.

UPDATE: I tried with the same credentials and on the same machine/virtualenv by using twython and got no problems here. Thus, it must relate to tweepy, but I don't know where to look further. I also found a post on StackOverflow with the same isue (http://stackoverflow.com/questions/27910383/tweepy-not-working-no-error-message)

joshthecoder:
Remove usage of urlencode_noplus and instead let Requests handle the encoding work for us.

Fixes #556

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -20,7 +20,7 @@ from tweepy.models import Status
 from tweepy.api import API
 from tweepy.error import TweepError
 
-from tweepy.utils import import_simplejson, urlencode_noplus
+from tweepy.utils import import_simplejson
 json = import_simplejson()
 
 STREAM_VERSION = '1.1'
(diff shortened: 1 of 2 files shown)


ohnorobo:
This fixes filtering in python3 for me. :+1:

PR #559, opened 2015-02-16
Streaming improvements

pvanderlinden:
This will fix some issues I encountered with the streaming api:
- exceptions are raise with a changed traceback
- when you stop and start the connection twitter will resume with gzip because of the headers
- callback for keep-alive to be able to detect lost connection

Also I should mention that the async implementation really slows down the throughput even with just a time.sleep in the other thread. I implemented it with gevent which works fine. Maybe I should raise a pull request which implements this?

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -79,6 +79,10 @@ class StreamListener(object):
         else:
             logging.error("Unknown message type: " + str(raw_data))
 
+    def keep_alive(self):
+        """Called when a keep-alive arrived"""
+        return
+
     def on_status(self, status):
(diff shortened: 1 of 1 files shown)


joshthecoder:
Looks good to me. I would be interested in adding a gevent example.

Do we need to make changes to Stream to support gevent or other evented I/O libraries?
I agree the threaded approach is probably not ideal for high performance use cases.

pvanderlinden:
@joshthecoder 
I will raise a PR with a gevent example soon. Just finishing of other things, but the code with gevent seems to work fine.

PR #561, opened 2015-02-18
Prevent infinite read loop in tweepy Stream

michaelbrooks:
I have noticed that my twitter streaming programs using tweepy (3.2.0 with requests 2.4.3) have been going into an infinite loop after a couple of days running. They seem to get stuck in the `ReadBuffer.read_line()` method. From breaking into it with a debugger, it looks like the underlying `raw` stream (in the `ReadBuffer` this is `self._stream`) is closed (`self._stream.closed` is True).

I am not certain how the stream gets closed, but probably there is a network glitch or Twitter just booted me randomly. I am surprised that no exception gets thrown from requests when you try to read from a closed response, but I don't know the library that well, so maybe this is intentional.

In the stream read loops in `stream.py`, there is currently only one check if the underlying stream has been "closed" (line 337), and it is actually *outside* the read loop. I looked back at the history of the file and found in #400 that *before* switching to the requests library, there were checks for `isclosed()` [all over the place](https://github.com/tweepy/tweepy/pull/400/files#diff-07400b70c336387bf35f0433991068f0L206), and after there was only the one.

Anyway, to try and prevent this from happening, I added checks to make sure the response stream is still open in the read loops within the `ReadBuffer` class, as well as in `Stream._read_loop()` method. I also updated line 337 at the bottom of the `_read_loop()` to use `resp.raw.closed` instead of `resp.raw._fp.closed()` since this seems like it is supposed to be private API.

I added a test in the `TweepyStreamReadBuffer` class that mocks a closed stream with no content available and then tries to call `read_line()` on it. Before I modified `stream.py` this test failed because `read_line()` looped infinitely. Now `read_line()` returns `None`, which hopefully is the desired result in this situation. Note that I had to do some extra work to block a true infinite loop inside the test, hence all the extra complexity there.

I don't know if I fixed this the best way, but the problem is pretty serious so I hope this can be fixed and released quickly. Thank you!

### DIFF
--- a/tests/test_streaming.py
+++ b/tests/test_streaming.py
@@ -162,12 +162,8 @@ class TweepyStreamReadBuffer(unittest.TestCase):
         # Mock it's read function so it can't be called too many times
         mock_read = MagicMock(side_effect=on_read)
 
-        # Add an _fp member to it, like a real requests.raw stream
-        mock_fp = MagicMock()
-        mock_fp.isclosed.return_value = True
-
         try:
-            with patch.multiple(stream, read=mock_read, _fp=mock_fp, create=True):
+            with patch.multiple(stream, create=True, read=mock_read, closed=True):
                 # Now the stream can't call 'read' more than call_limit times
                 # and it looks like a requests stream that is closed
                 buf = ReadBuffer(stream, 50)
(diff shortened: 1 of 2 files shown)


joshthecoder:
Looks good to me. :+1:  Thanks!

PR #562, opened 2015-02-21
Update api.py

RussellTaylor83:
Made catch statement error message more explicit for _pack_image. Now outputs  

tweepy.error.TweepError: Unable to access file: No such file or directory


joshthecoder:
:+1:  Thanks for the improvement!

PR #566, opened 2015-02-23
Streaming fixes

pvanderlinden:
Some small fixes:
- rearrange the methods on the listener to categorize them by non-data and data callbacks
- moved the closed callback to the listener
- some cleanup
- support for compressed http streams
- handle disconnects properly

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -40,6 +40,22 @@ class StreamListener(object):
         """
         pass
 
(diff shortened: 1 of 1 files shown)


pvanderlinden:
The gzip compression might give some problems with the chunk size implementation

pvanderlinden:
Using the blocking httplib (which is used by requests -> urllib3 -> httplib), is not going to make this work.

PR #578, opened 2015-03-20
update_status: first positional argument should be 'status'

### LINKED ISSUE #554: Api.update_status broken
api.update_status('Testr')

results in error code 400

api.update_status(status='Test')
posts the status correctly

wjt:
Passing the tweet text as the lone positional argument to update_status:

     api.update_status(u'Hi mum!')

is suggested in much documentation and in several tests, but f99b1da
introduced media_ids as the first argument, breaking this API.

This change will break anyone who uses tweepy >= 3.2 and assumes they
can pass media_ids as the first positional; but I expect that to be a
much smaller set than people who have followed the example in the docs,
who find their application mysteriously fails to post tweets after
upgrading tweepy.

Fixes #554 

(I haven't successfully run the tests, I'm hoping Travis will do that for me and save me the hassle of working out the token situation.)

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -87,6 +87,17 @@ class TweepyAPITests(TweepyTestCase):
         deleted = self.api.destroy_status(id=update.id)
         self.assertEqual(deleted.id, update.id)
 
+    @tape.use_cassette('testupdateanddestroystatus.json')
+    def testupdateanddestroystatuswithoutkwarg(self):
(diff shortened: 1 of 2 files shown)


cyphar:
Is there a reason this isn't being merged? This fixes a badly breaking change for a whole bunch of different things.

joshthecoder:
Looks good :+1:

PR #588, opened 2015-04-04
API.update_status() problem in example code

lackofdream:
It seems that the API.update_status() function has changed however the example code hasn't.
And It causes a 400 Bad Request.


joshthecoder:
:+1:  Thanks

PR #596, opened 2015-04-21
Adds API for account/settings

kimtree:
I recently find that there's no account setting API implements in tweepy.

Twitter is trying to add more options in account setting so we should prepare for that changes.


joshthecoder:
Looks good to me. Thanks!

PR #606, opened 2015-05-10
Streaming: don't decode stream bytes until json.decode

7diracseas:
Some people (including me) seem to be having issues with UTF-8 encoded tweets when reading a stream. This *should* fix it.

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -150,7 +150,7 @@ class ReadBuffer(object):
 
     def __init__(self, stream, chunk_size):
         self._stream = stream
-        self._buffer = u""
+        self._buffer = ''
         self._chunk_size = chunk_size
 
     def read_len(self, length):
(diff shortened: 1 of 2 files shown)


tweepy/streaming.py
```diff
@@ -158,7 +158,7 @@ def read_len(self, length):
             if len(self._buffer) >= length:
                 return self._pop(length)
             read_len = max(self._chunk_size, length - len(self._buffer))
-            self._buffer += self._stream.read(read_len).decode("ascii")
```
7diracseas: Not really sure why we have to decode as ascii here.

tweepy/streaming.py
```diff
@@ -150,7 +150,7 @@ class ReadBuffer(object):
 
     def __init__(self, stream, chunk_size):
         self._stream = stream
-        self._buffer = u""
+        self._buffer = ''
```
7diracseas: Keep buffer as bytes until we're ready to decode from JSON.

7diracseas:
A lot of the tests for the streaming functionality are out of date. I can spend some time fixing those tests as well.

joshthecoder:
Looks good to me. :+1:  Thanks!

PR #633, opened 2015-07-18
Parse quoted_status dict to Status object

Pietro210:
The quoted_status property of a Tweet is not currently being parsed but it is stored as a simple dict.

I think it would better to be consistent and convert it to a Status object just like retweeted_status.

### DIFF
--- a/tweepy/models.py
+++ b/tweepy/models.py
@@ -93,6 +93,8 @@ class Status(Model):
                     setattr(status, 'source_url', None)
             elif k == 'retweeted_status':
                 setattr(status, k, Status.parse(api, v))
+            elif k == 'quoted_status':
+                setattr(status, k, Status.parse(api, v))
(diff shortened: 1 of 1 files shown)


dgw:
Came here to open an issue about this and, what do you know, there's already an open PR to fix the problem I've been having! Who do I talk to about getting this merged into the next release? Right now I'm facing writing a second version of my tweet handling function that would work on the dict form, and I'd really rather not when the solution is so simple.

dgw:
I tested this the hacky way (edited the source in `site-packages`, deleted the generated `.pyc` file, and restarted the Python app in question) and it works.

Actually I just added `or k == 'quoted_status'` to the existing `elif` for retweets because it's more concise, but the effect is the same.

joshthecoder:
:+1:  Thanks for the contribution!

PR #635, opened 2015-07-23
Handle content-type header charset value for streaming API

### LINKED ISSUE #615: TypeError: Can't convert 'bytes' object to str implicitly
I was following the streaming api tutorial and got this error. Other threads say it has been fixed.
Using master branch code.

typeerror

varzac:
I tested on python 3.4 and python 2.7.

I was unable to get the tox tests properly running on my machine.  This should fix #615 properly as well as properly use the content-type charset http header if it is present.

Closes #615

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -7,6 +7,7 @@
 from __future__ import absolute_import, print_function
 
 import logging
+import re
 import requests
 from requests.exceptions import Timeout
 from threading import Thread
@@ -148,10 +149,11 @@ class ReadBuffer(object):
     use small chunks so it can read the length and the tweet in 2 read calls.
     """
 
-    def __init__(self, stream, chunk_size):
+    def __init__(self, stream, chunk_size, encoding='utf-8'):
         self._stream = stream
-        self._buffer = ''
+        self._buffer = six.b('')
         self._chunk_size = chunk_size
+        self._encoding = encoding
 
     def read_len(self, length):
         while not self._stream.closed:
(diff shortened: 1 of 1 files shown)


swstack:
If you put "closes #615"  in your PR body it will close the issue automagically when this is merged.

tweepy/streaming.py
```diff
@@ -168,7 +170,7 @@ def read_line(self, sep='\n'):
                 return self._pop(loc + len(sep))
             else:
                 start = len(self._buffer)
-            self._buffer += self._stream.read(self._chunk_size)
+            self._buffer += self._stream.read(self._chunk_size).decode(self._encoding)
```
posborne: I'm not sure it is safe to perform decoding on arbitrary chunks of data as they come in as you could end up in a situation where you are using a valid decode on incomplete data.  Take for instance the following (python3.4):

```
>>> skull = "☠"
>>> skull_encoded = skull.encode('utf-8')
>>> skull_encoded
b'\xe2\x98\xa0'
>>> skull_encoded[:-1].decode('utf-8')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 0-1: unexpected end of data
```

varzac: That is a good point and one that I didn't really consider.  How do you feel about doing the decode as a part of the pop function.  I.e. keep everything internal to the ReadBuffer as bytes, but decode on returns.  The other option of course is to just keep everything in bytes and return bytes, and just offload the decoding to the person using ReadBuffer.  What is your opinion?

posborne: I think doing it on pop probably makes sense.

danielquinn:
:+1:

PR #636, opened 2015-07-26
make update_status example work better

techieshark:
I was getting the error mentioned in [this stack overflow question](https://stackoverflow.com/questions/28527352/traceback-when-updating-status-on-twitter-via-tweepy) and found that [this answer](http://stackoverflow.com/a/28527452/1024811) solved the problem.


wjt:
Rendered unnecessary by #578 (which was just merged).

techieshark:
Super! Thanks!

PR #637, opened 2015-07-27
Fix streaming.py to run on BOTH PY3 and py2

7kry:
Previously I requested to merge this commit (https://github.com/tweepy/tweepy/commit/48d2744d7b948f166ebf67796d7fec4c278688a1).
Indeed this is not a good solution, but after this commit (https://github.com/tweepy/tweepy/commit/2bbef250ffaab8297c6b35e3225bfef28a5e65e5), this function may have NOT been able to use with Python 3.

HAVE YOU EVER TESTED WITH PYTHON 3 ENVIRONMENT????

Now then, I made some fixes to enable the code to be used with both py3 and py2.
I tested with this code (https://gist.github.com/7kry/27459f70cfc3745fe068/69a088ddaa670d7f5f8938c2b4cff59158dd7e68).
Please consider to merge.

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -46,7 +46,10 @@ class StreamListener(object):
         Override this method if you wish to manually handle
         the stream data. Return False to stop stream and close connection.
         """
-        data = json.loads(raw_data)
+        if six.PY2:
+          data = json.loads(raw_data)
+        elif six.PY3:
+          data = json.loads(raw_data.decode('ascii'))
 
         if 'in_reply_to_status_id' in data:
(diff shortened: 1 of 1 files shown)


varzac:
I have a pull request aimed at fixing the same issue, but I solved it in a slightly different way.  See #635 I use the content-type header instead of assuming an ascii decode, and default to utf-8 instead of ascii if one isn't present. I believe the default JSON encoding type is utf-8 not ascii.

7kry:
> The individual messages streamed by this API are JSON encoded.
https://dev.twitter.com/streaming/overview/processing

PR #655, opened 2015-09-28
Video Upload

Choko256:
Make tweepy handle MP4 videos upload according to the Twitter API documentation for chunked uploads :
https://dev.twitter.com/rest/public/uploading-media#chunkedupload

### DIFF
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -8,10 +8,11 @@ import os
 import mimetypes
 
 import six
+import urllib
 
 from tweepy.binder import bind_api
 from tweepy.error import TweepError
-from tweepy.parsers import ModelParser, Parser
+from tweepy.parsers import ModelParser, Parser, RawParser
 from tweepy.utils import list_to_csv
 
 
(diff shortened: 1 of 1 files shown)


obskyr:
I was going to work on this myself, but as it turns out, someone already had. Nice!

This doesn't seem to be usable, though. Sure, this allows for uploading videos, but there's no way to actually tweet them. The `update_with_media` (or really the `update_status` method, since `update_with_media` [is deprecated](https://dev.twitter.com/rest/reference/post/statuses/update_with_media) needs to be able to handle video files / filenames for it to actually be usable, right?

Choko256:
Hello @obskyr !
No need to update `update_status` function. You can send the video link in the media_ids argument and it rocks :+1: 
I tried this last week and it works really fine ! Of course you will have to send Twitter-compliant videos (MP4 <= 30sec, 15MB max, 1280x1024max, etc. you can have the list of constraints in the given link). Otherwise, a TweepError is raised.

I did not change anything else and it works fine for me :)
Let me know if something goes wrong ;)

Choko256:
Tip: You cannot send videos *and* images in a tweet. You can only send 1 video **or** 4 images max.

obskyr:
Oh, the docs don't mention the `media_ids` parameter. The docs are out of date in general, in fact. Cool, so there is a way to use it.

Regardless, don't you think it should be easier to tweet videos with the API? It'd be way more dev-friendly if you could just tweet a video with a filename / file object and it'd handle the upload under the hood for you, like what is done with images. All that would be needed would be some checks to see whether a file is a video or an image, which is easy (basically just check the extension).

Choko256:
Yes you're right !
I will create a wrapper function to send directly tweets with medias by handling upload automatically ;)

Thanks!

askaliuk:
Thank you for awesome work, @Choko256.
@joshthecoder do you plan to merge it?

jeremylow:
Hey, great work. I was just trying to find a solution to a GIF problem and ended up here. One thing that I found (since my code is python3): you're going to run into problems with the urllib import statement and the supported versions of python. Maybe change from:

    import urllib

to:

    if six.PY2:
        from urllib import urlencode
    elif six.PY3:
        from urllib.parse import urlencode

And then change lines 1444 & 1481 to just use ``urlencode``.

askaliuk:
Fails sometimes for me with following exception:
```
TweepError: {"request":"\/1.1\/media\/upload.json","error":"SegmentIndex must be between 0 and 999."}
  File "tweepy/api.py", line 252, in video_upload
    )(*args, **kwargs)
  File "tweepy/binder.py", line 243, in _call
    return method.execute()
  File "tweepy/binder.py", line 227, in execute
    raise TweepError(error_msg, resp)
```

jeremylow:
How large is the file you're trying to upload

askaliuk:
Found root cause:
https://github.com/tweepy/tweepy/pull/655/files#diff-ea5dd38a4efd9ff36c96e04ab0597cfbR239

```
File size = 5219933.
chunk_size = 4096
nloops = 1275
```

I think the code should raise an exception if nloops > 999 and don't try to upload. Also explain to the method caller, that he/she needs to increase `chunk_size` or decrease `file size`.

jeremylow:
4k is too small as a default. Should probably be set to something more reasonable like 1MB. 

Rather than returning an error, if the user really wants a small chunk size, disallow any value smaller than 16K, so that it always succeeds. (Max size / 16K = 960 chunks)

fitnr:
This is great and I would love to see it in Tweepy asap.

But, what do you think about having a single media_upload function that's agnostic about file type? I think it would just need to have a single `mimetypes.guess_type` check, and then send the sile to the correct function. This would save trouble of having to check file file for applications that might be upoading video and images.

jason-hyland-heyhuman:
Hi,

Just getting started with python and the twitter APi for a twitter bot, can anyone help me with where I might be going wrong with the video send:

twitterApi.update_status(filename = 'MP4_SIZE_TEST_2.mp4',status=replyText, in_reply_to_status_id=tweetId, file = 'MP4_SIZE_TEST_2.mp4')

Im simply trying to reply to an @mention tweet with a video and a message (message is created earlier and assigned to the replyText var).

I get the message ok, and works fine if i try with a message and an image, but video never makes it through or throws error

beaumartinez:
I'd be very happy to see Tweepy support this. Any updates?

waylander47:
Hi guys,

Did someone find a solution ?
Because I've got the same issue, when I post a video I got this answer from twitter : {"media_id":55783223323555,"media_id_string":"555556565465656","size":323161,"expires_after_secs":86400,"video":{"video_type":"video\/mp4"}}
Like if everything went fine but there is no video on Twitter.

Do you have any idea ?

Best regards.

waylander47:
My bad everything fine we just need to update status with the media ids.

tweepy/api.py
```diff
@@ -211,6 +212,60 @@ def media_upload(self, filename, *args, **kwargs):
             upload_api=True
         )(*args, **kwargs)
 
+    def video_upload(self, filename, *args, **kwargs):
+        """ :reference https://dev.twitter.com/rest/reference/post/media/upload-chunked
```
AgoraSecurity: I think this line should be: https://developer.twitter.com/en/docs/media/upload-media/uploading-media/chunked-media-upload

Harmon758:
Superseded by #672 and #929 

@jason-hyland-heyhuman [`API.update_status`](https://tweepy.readthedocs.io/en/latest/api.html#API.update_status) does not accept `filename` or `file` parameters, and those aren't being added by this PR.

marvic2409:
Any updates on this?

Harmon758:
The video-upload branch / pull request #1486 should be complete now.
Any feedback or review would be appreciated.

PR #664, opened 2015-11-01
Allow 'full_text' param when getting direct messages

tjphopkins:


tjphopkins:
Twitter have removed the character length limit from direct messages so that they can now be up to 10k chars. Without sending the param full_text=true when retrieving messages, long messages will be returned truncated.

PR #675, opened 2015-11-22
Fixes stream timeouts on shotty connections

### LINKED ISSUE #617: Timeouts and Aborted Connections after running code for long times
I am using Twitter API to scan through tweets and users, but because of Twitter's API limits, scanning thousands of users/tweets can take hours. Fine. However, when I run Twitter API code through Python using Tweepy, after a few hours of the code running the code gets interrupted and I get one of two error messages:

tweepy.error.TweepError: Failed to send request: ('Connection aborted.', error(54, 'Connection reset by peer'))

Another error:

tweepy.error.TweepError: Failed to send request: HTTPSConnectionPool(host='api.twitter.com', port=443): Read timed out. (read timeout=60)

It seems Twitter has a timeout limit that stops access after a certain point? Or not? Is this a Twitter API issue, or a Tweepy one?

This doesn't happen for small sizes, only when the code has to spend more than about an hour running. Of course one sees the problem with walking away from script that should take hours to run, only to come back and find that it's aborted partway through.

No one seems to know why this happens. Posted on several forums and spoken to several people..Could be a bug.

ccphillippi:
Now catches ConnectionError, ReadTimeOutError, ProtocolError in addition to Timeout, ssl.SSLError during stream, handling as a timeout to resolve #617 per suggestion by @jcapde87

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -9,7 +9,8 @@ from __future__ import absolute_import, print_function
 import logging
 import re
 import requests
-from requests.exceptions import Timeout
+from requests.exceptions import Timeout, ConnectionError
+from requests.packages.urllib3.exceptions import ReadTimeoutError, ProtocolError
 from threading import Thread
 from time import sleep
 
@@ -261,7 +262,7 @@ class Stream(object):
                     self.snooze_time = self.snooze_time_step
                     self.listener.on_connect()
                     self._read_loop(resp)
-            except (Timeout, ssl.SSLError) as exc:
+            except (Timeout, ssl.SSLError, ConnectionError, ReadTimeoutError, ProtocolError) as exc:
                 # This is still necessary, as a SSLError can actually be
                 # thrown when using Requests
                 # If it's not time out treat it like any other exception


ZeerakW:
Is there any reason this hasn't been pulled?

sam-s:
this seems to be an improvement, but not a complete fix.
I still see 
```
TweepError: Failed to send request: ('Connection aborted.', error(54, 'Connection reset by peer'))
```

joshthecoder:
Closing this is pretty old now and not active

Harmon758:
This is not stale.

tweepy/streaming.py
```diff
@@ -261,7 +262,7 @@ def _run(self):
                     self.snooze_time = self.snooze_time_step
                     self.listener.on_connect()
                     self._read_loop(resp)
-            except (Timeout, ssl.SSLError) as exc:
+            except (Timeout, ssl.SSLError, ConnectionError, ReadTimeoutError, ProtocolError) as exc:
```
Harmon758, line 265:
```suggestion
            except (requests.ConnectionError, requests.Timeout, ssl.SSLError,
                    urllib3.exceptions.ReadTimeoutError,
                    urllib3.exceptions.ProtocolError) as exc:
```

tweepy/streaming.py
```diff
@@ -9,7 +9,8 @@
 import logging
 import re
 import requests
-from requests.exceptions import Timeout
+from requests.exceptions import Timeout, ConnectionError
+from requests.packages.urllib3.exceptions import ReadTimeoutError, ProtocolError
```
Harmon758, line 13:
```suggestion
import urllib3
```

Harmon758 (changes requested):
I think consistently including an explicit reference to the corresponding library/package for each exception handled is better here, so that it's clearer which exceptions are being handled.

Also, the StreamListener method should probably be updated to indicate that it could be called due to connection errors as well as timeouts.

Harmon758:
I've gone ahead and resolved this with https://github.com/tweepy/tweepy/commit/68e19cc6b9b23d72369ca1520093770eb18a5a9f (and https://github.com/tweepy/tweepy/commit/8f622971f2a36f49ad54b70cf4f98464487a4716).

dheeman00:
Not sure why this thread is close. But this issue still exists with the tweepy library but there is not any proper solution to handle this issue.

Harmon758:
As the closing comment says, this was resolved with https://github.com/tweepy/tweepy/commit/68e19cc6b9b23d72369ca1520093770eb18a5a9f.
Feel free to create a new issue if you can still reproduce it with the latest development version on the master branch.

dheeman00:
@Harmon758 Thanks for your response. I am trying to extract the follower IDs for the top Twitter followers. But I encounter this error if there is any issue with the connection (either from the firewall or server-side). But can the code be modified in a way where the extraction starts from the point it stopped? I am not willing to crawl from the beginning of the list.

Harmon758:
That doesn't seem like it has anything to do with this PR, which specifically involves connection issues while streaming.
Feel free to create a discussion with your relevant code if you have questions about using Tweepy.
Without seeing the actual code (and traceback), it's not really possible to tell you how it could be modified to handle your issue.

dheeman00:
@Harmon758 I will go ahead and create a new thread for my issue.

PR #676, opened 2015-11-22
Allow user defined rate limit notifications.

nawatts:
Allow a callable to be passed to `wait_on_rate_limit_notify` so that a rate limit notice can trigger actions other than printing to stdout.


ZeerakW:
Oh this is a neat idea!

PR #688, opened 2015-12-24
Fixed typo in streaming_how_to.rst

m4rc1e:
We should be calling the instance of MyStreamListener(), which is myStreamListener. Not the object itself.


joshthecoder:
Dup of #695

PR #700, opened 2016-02-05
Update doc version in README

remram44:
You should also probably change the default version on read the docs; currently http://docs.tweepy.org/ gets you to 3.2.0.


joshthecoder:
Whoops. I always forget to bump that one. Thanks.

PR #708, opened 2016-03-07
add missing streaming.py link

snitzr:
links streaming.py file from `examples` directory

### DIFF
--- a/docs/streaming_how_to.rst
+++ b/docs/streaming_how_to.rst
@@ -1,4 +1,4 @@
-.. _streaming_how_to:
+.. _streaming.py: https://github.com/tweepy/tweepy/blob/master/examples/streaming.py
 .. _Twitter Streaming API Documentation: https://dev.twitter.com/streaming/overview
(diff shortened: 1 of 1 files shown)


docs/streaming_how_to.rst
```diff
@@ -1,4 +1,4 @@
-.. _streaming_how_to:
+.. _streaming.py: https://github.com/tweepy/tweepy/blob/master/examples/streaming.py
```
Harmon758: I believe the `streaming.py` file being referenced is the [source code file](https://github.com/tweepy/tweepy/blob/master/tweepy/streaming.py) and not the [example](https://github.com/tweepy/tweepy/blob/master/examples/streaming.py).
Also, for consistency, this should be a separate new link without replacing the title link.
```suggestion
.. _streaming_how_to:
.. _streaming.py: https://github.com/tweepy/tweepy/blob/master/tweepy/streaming.py
```

snitzr:
PR contained error in reference as Harmon758 described.

PR #710, opened 2016-03-08
Fixing #698

runn1ng:
Fixing #698 by the fix proposed there

(I don't really know what this is doing or why, I just copy-pasted the suggested fix by @rxcomm )


runn1ng:
Seems like a duplicate of 
https://github.com/tweepy/tweepy/pull/694

and seems like nobody is actually merging back pull requests in this project. Which is sad kinda

PR #718, opened 2016-04-04
Support compressed streaming

mannby:
Based on example from https://rationalpie.wordpress.com/2010/06/02/python-streaming-gzip-decompression/


maesfreek:
Can this pull request be merged? Seems to be a good change

Harmon758:
This shouldn't be necessary, as the requests library by default accepts gzip encoding and decompresses it automatically for each response.

PR #719, opened 2016-04-05
py3-ify getting_started.rst

nutztherookie:
.. also still compatible with python2


nutztherookie:
I realize this would have to be applied to a lot more files. I can re-create a pull-request with everything updated (later.. at some point)

Issue #716, opened 2016-03-29
media metadata: alt text support


fitnr:
Twitter has added a media metadata API endpoint. Currently it supports adding alt text to an image: https://blog.twitter.com/2016/alt-text-support-for-twitter-cards-and-the-rest-api

Other parts of the API are also being extended with a `include_ext_alt_text` parameter.

The above link notes that "Full documentation for this endpoint will soon be published and linked directly once available".

tinruufu:
[it lives](https://dev.twitter.com/rest/reference/post/media/metadata/create) and it would be very useful for making my bots more accessible

Hyperclaw79:
Please merge this. It's a nice commit.

arizonagroovejet:
Hello to anyone who, like me, has searched the tweepy documentation for "alt_text", got zero results, then arrived here from a Google search. Maybe, particularly if like me you have little idea what you're doing with python, you'll be interested in this code I just wrote which tweets multiple images with the same alt_text using a modified version of what @fitnr wrote to add alt_text support to tweepy two years ago. 
```
#!/usr/bin/python

import tweepy
import sys
import os

consumer_key='whatever'
consumer_secret='whatever'
access_token='whatever'
access_token_secret='whatever'

auth = tweepy.OAuthHandler(consumer_key, consumer_secret)
auth.set_access_token(access_token, access_token_secret)

api = tweepy.API(auth)

# adapted from https://github.com/fitnr/tweepy/commit/b387331c174a451cb8dba44b4e0c7988a92bad1b
def media_metadata_create(self, media_id, alt_text):
  post_data = {
   "media_id": media_id,
   "alt_text": {
     "text": alt_text
    }
  }
  json = tweepy.utils.import_simplejson()
  return tweepy.binder.bind_api(
    api=self,
    path='/media/metadata/create.json',
    method='POST',
    allowed_param=[],
    require_auth=True,
    upload_api=True
  )(post_data=json.dumps(post_data))

media_ids=[]

# arvg[2] is a directory containing images to tweet
for filename in sorted(os.listdir(sys.argv[2])):
  media_ids.append(api.media_upload(filename).media_id_string)

# argv[3] is alt text
for m in media_ids:
  media_metadata_create(api, m, sys.argv[3])

# arg[1] is the status text
api.update_status(status=sys.argv[1],media_ids=media_ids)
```

PR #727, opened 2016-05-06
add media_metadata_create method to tweepy.API

fitnr:
This pull adds support for the `media/metadata/create` endpoint (issue #716). I found that the API wanted the JSON as a string, hence the `json.dumps` call. If there's another way to accomplish this I would be happy to hear about it.


edent:
This works perfectly. Any idea when it'll be merged?

PR #733, opened 2016-05-20
Update api.rst

andjelx:
Added note how to properly call  **send_direct_message**


Harmon758:
This is not specific to `API.send_direct_message` and shouldn't go here.
Feel free to create a new PR with a general explanation of the format of documentation for parameters in a more appropriate location.

PR #755, opened 2016-07-10
Added comment about process_status

Parth-Vader:
Added "# process_status function should be defined above." comment so that new users could know that process_status has to be defined by themselves.


Harmon758:
I don't really think this is needed in the documentation and might instead introduce confusion.
There's already comments indicating that the status should be processed inside the loop.
The `process_status` method being used is simply an example and isn't necessary.

PR #766, opened 2016-07-26
Fix bug in JSONParser.parse() which causes it to return None

### LINKED ISSUE #765: JSONParser.parse() returns None if only 'next_cursor' available.
It looks like someone has broken an if/else when refactoring, which causes python to drop off the end of the function without returning anything useful.

https://github.com/tweepy/tweepy/blame/master/tweepy/parsers.py#L57

alsuren:
Re-combine tower-of-ifs into a single if/else.

Fixes #765

### DIFF
--- a/tweepy/parsers.py
+++ b/tweepy/parsers.py
@@ -54,11 +54,11 @@ class JSONParser(Parser):
             raise TweepError('Failed to parse JSON payload: %s' % e)
 
         needs_cursors = 'cursor' in method.session.params
-        if needs_cursors and isinstance(json, dict):
-            if 'previous_cursor' in json:
-                if 'next_cursor' in json:
-                    cursors = json['previous_cursor'], json['next_cursor']
-                    return json, cursors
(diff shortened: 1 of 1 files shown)


tweepy/parsers.py
```diff
@@ -54,11 +54,11 @@ def parse(self, method, payload):
...
-        if needs_cursors and isinstance(json, dict):
-            if 'previous_cursor' in json:
-                if 'next_cursor' in json:
-                    cursors = json['previous_cursor'], json['next_cursor']
-                    return json, cursors
+        if needs_cursors and isinstance(json, dict) \
+                and 'previous_cursor' in json \
+                and 'next_cursor' in json:
+            cursors = json['previous_cursor'], json['next_cursor']
+            return json, cursors
```
lanzz: This will still not return any cursors if only `next_cursor` is present in the response. Looks like it should return `json, (None, next_cursor)` if there's only one?

PR #781, opened 2016-09-13
Status and User Models example

Ranlevi:
Hi, 
I'm a noob to Open Source development, so please excuse any silly mistakes on my part.
I've created an example which I think might be useful for users:  This example explains tweepy's Status & User Models, and demonstratesparsing the info found in a single tweet ('Status').

Best Regards,
Ran

### DIFF
--- a/examples/Status_And_User_Models.py
+++ b/examples/Status_And_User_Models.py
@@ -1,0 +1,80 @@
+import tweepy
+
+#   This example explains tweepy's Status & User Models, and demonstrates
+#   parsing the info found in a single tweet ('Status')
+#
+# Author: Ran Levi, @ranlevi, ran@cmpod.net
+
+#For this example, We use a particular tweet with the following ID:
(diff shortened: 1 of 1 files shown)


Harmon758:
I'm going to close this as there's numerous issues with the example, it's outdated now, and I'm not sure what the purpose of the example is in recreating a Status object that already exists.
If those are valid credentials, you're going to want to revoke and regenerate them, and they shouldn't have been included in the example and PR.
The formatting leaves a lot to be desired. See [PEP 8](https://www.python.org/dev/peps/pep-0008/).
`'Text'` (and `'Id'`) will never be a valid attribute and the other `attributes` should always exist.
`contributors_enabled` has been deprecated and the documentation links are now outdated.
[`API.get_status`](https://tweepy.readthedocs.io/en/latest/api.html#API.get_status) should have been used to get a single Status rather than `API.statuses_lookup`.

PR #797, opened 2016-10-11
Fix fatal errors in streaming

feinsteinben:
Trying to fix several fatal errors, occurring when the connection breaks while reading. These errors were observed in two locations:
- buf.read_line() or buf.read_len() returns None when the connection is closed before completing the read. later  on, this output (None) was considered as a string and the program crashed as a side-effect (with AttributionError).
- httplib.IncompleteRead was caught in requests, and propagated through buf.read_line() or buf.read_len(). In this case, the retry mechanism didn't catch HTTP errors so it didn't try to reconnect as expected.

These problems were fixed as follows:
- buf.read_line() or buf.read_len() fails with httplib.IncompleteRead when the connection breaks in the middle.
- The retry mechanism treats httplib.HTTPError similarly to timeouts and disconnections errors, following [Twitter's reconnecting advice](https://dev.twitter.com/streaming/overview/connecting).

### DIFF
--- a/tests/test_streaming.py
+++ b/tests/test_streaming.py
@@ -175,6 +175,23 @@ class TweepyStreamReadBufferTests(unittest.TestCase):
         # The mocked function not have been called at all since the stream looks closed
         self.assertEqual(mock_read.call_count, 0)
 
+    @skip
+    def test_read_incomplete_buffer(self):
+        from six.moves import http_client as httplib
+        for length in [1, 2, 5, 10, 20, 50]:
+            with self.assertRaises(httplib.IncompleteRead):
+                msg = "11\n".replace('\n', '')
+                buf = ReadBuffer(six.BytesIO(six.b(msg)), length)
+                buf.read_line()
+            with self.assertRaises(httplib.IncompleteRead):
+                msg = '{id:12345}'
+                buf = ReadBuffer(six.BytesIO(six.b(msg)), length)
+                buf.read_len(len(msg) - 1)
+            with self.assertRaises(httplib.IncompleteRead):
+                msg = '{id:23456, test:"blah"}\n'
+                buf = ReadBuffer(six.BytesIO(six.b(msg)), length)
+                buf.read_len(len(msg) - 2)
+
     def test_read_unicode_tweet(self):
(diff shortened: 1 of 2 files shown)


feinsteinben:
#576 
#595 
#734 
#780

feinsteinben:
Resolves #576

feinsteinben:
Hi @edgardojosec,

Did you run your code before applying this patch? did it work as planned?

Anyway, I can think of one change that may solve this issue. I'll push it shortly.

Cheers,
Ben

feinsteinben:
Now it also fixes another error, when requests.packages.urllib3.exception.ReadTimeout is raised. requests doesn't re-raise it as Timeout, so our retry mechanism didn't catch it.

With the above fixes (excluding the last one), tweepy ran continuously for a whole week without crashing.

feinsteinben:
Update: after applying this fix, my app runs for the 6 week in a raw (continuously) without any trouble.

tests/test_streaming.py
```diff
@@ -175,6 +175,22 @@ def on_read(chunk_size):
         # The mocked function not have been called at all since the stream looks closed
         self.assertEqual(mock_read.call_count, 0)
 
+    def test_read_incomplete_buffer(self):
+        from six.moves import http_client as httplib
+        for length in [1, 2, 5, 10, 20, 50]:
+            with self.assertRaises(httplib.IncompleteRead):
+                msg = "11\n".reaplce('\n', '')
```
DeanF: You meant `replace`. Seems like the file isn't in `travis-tests.cfg` so this file isn't run.

feinsteinben: Thanks. Fixed.

feinsteinben: @DeanF test_streaming.py runs endlessly, even without my tests. Are you familiar with it?

Anyway, thanks for reviewing my pr :-)

alexanderkoumis:
These incomplete reads are still a problem, why hasn't this been merged?

Harmon758:
This is stale now and should be resolved with https://github.com/tweepy/tweepy/commit/68e19cc6b9b23d72369ca1520093770eb18a5a9f and https://github.com/tweepy/tweepy/commit/37e701e2193dcf617e7235d3b004f790c5ff784e.

PR #799, opened 2016-10-13
Update in the Documentation about 'oauth_token' 

vatsalparekh:
Due to maybe old docs, #794  was occurring in `get_access_token( )`


korsunowk:
Thanks!

Issue #598, opened 2015-04-22
statuses_lookup with map=True fails


geosoco:
statuses_lookup calls bind_api with payload_list=True. This works until you call statuses_lookup with map=True. The resulting format from Twitter is an object with a single key ('id') and the value is an object with each id from the request being a key. Values are then the statuses.

```
{ 'id': { '<tweet id>': <status> , '<tweet id>': None, '<tweet id>': <status>, ... }
```

I'm not super familiar with the code, but it looks like the payload method needs to be more than just the  'payload_list' flag.  

```
Traceback:
  File ".../dev/projects/tweepy/tweepy/api.py", line 99, in statuses_lookup
    trim_user, map_)
  File ".../dev/projects/tweepy/tweepy/binder.py", line 239, in _call
    return method.execute()
  File ".../dev/projects/tweepy/tweepy/binder.py", line 226, in execute
    result = self.parser.parse(self, resp.text)
  File ".../dev/projects/tweepy/tweepy/parsers.py", line 95, in parse
    result = model.parse_list(method.api, json)
  File ".../dev/projects/tweepy/tweepy/models.py", line 65, in parse_list
    results.append(cls.parse(api, obj))
  File ".../dev/projects/tweepy/tweepy/models.py", line 79, in parse
    for k, v in json.items():
AttributeError: 'unicode' object has no attribute 'items'
```

commadelimited:
I can confirm this behaviour.

commadelimited:
Okay. I've spent the better part of the day tracking this down. Here's what I can offer.

I'm calling the `statuses_lookup` method with the following signature:

```
tweet_ids = [638842021990518784, 638748404827443201]
# The first tweet id is an intentionally deleted id, while the second one is active
response = twitter_api.statuses_lookup(id_=tweet_ids, map_=True)
```

I've tracked it down to `/tweepy/parsers.py:line 80` which is currently the `parse` method of the `ModelParser`. I've included below what the value of the payload argument is for me, based on the above tweets. You can clearly see in Twitter's response that the result for 638842021990518784 is null which follows Twitter's documentation for responses with `map=true`.

```
[...]

PR #816, opened 2016-11-27
Properly handle data coming back if map_=True in statuses_lookup method

benhoff:
Fixes https://github.com/tweepy/tweepy/issues/598

I was going to write a test case for it, but I couldn't get the tests to work on my local machine and didn't see any documentation for getting tests up and running.

### DIFF
--- a/tweepy/models.py
+++ b/tweepy/models.py
@@ -60,6 +60,8 @@ class Model(object):
             a result set of model instances.
         """
         results = ResultSet()
+        if 'id' in json_list:
+            json_list = json_list['id']
         for obj in json_list:
             if obj:
                 results.append(cls.parse(api, obj))


benhoff:
The original issue is that if the map_ variable is set to true the data comes back in the from of 
 
    {'id': {TWEET_DATA}}

instead of the normal parsing which is just dictionary of tweet data. The `parse_list` method tries to call `items` on a string and chokes and dies since there is no `items` method on strings.

Weird corner case. Note that the `map_` keyword is supposed to turn back null tweet values, so I'm not sure what the resulting status models would look like.

benhoff:
@nikitasalkar1997 do you have to have `map_` set to True? Just remove that argument.

Harmon758:
This PR does not resolve #598. In fact, the exact same error will be raised, since as you and #598 outline, the single value of the dictionary you access with the key `"id"` is another dictionary, with string IDs as keys.
With this PR, `Model.parse_list` will iterate over those string IDs to pass to `Status.parse` and again raise the same error.

PR #818, opened 2016-12-04
added and implemented retweet_text() and updated requirements.txt

Theyflower:
see docstring for information about what it does

added mechanicalsoup to requirements.txt because retweet_text uses mechanicalsoup

some context for why I made this:
was working on a twitter bot at a hackathon and we wanted to get retweet text of a retweet, the api at that point had the capability to get the url of the retweet but we wanted to get the text of the original tweet that was being retweeted.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ requests>=2.11.1
 requests_oauthlib>=0.7.0
 six>=1.10.0
 PySocks>=1.5.7
+mechanicalsoup>=0.6.0
--- a/tweepy/utils.py
+++ b/tweepy/utils.py
@@ -10,7 +10,7 @@ import six
 from six.moves.urllib.parse import quote
 
 from email.utils import parsedate
-
+import mechanicalsoup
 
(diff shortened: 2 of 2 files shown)


fitnr:
This seems like its out of the scope of the library, which can already do this (although it takes two requests).

Harmon758:
This can actually already be accomplished with a single existing request using [`API.get_status`](https://tweepy.readthedocs.io/en/latest/api.html#API.get_status).
Using the status ID from the link, this will return a Status/[Tweet object](https://developer.twitter.com/en/docs/tweets/data-dictionary/overview/tweet-object.html).
The `retweeted_status` attribute (if it exists) of this Status object is actually another Status object.
This means you can simply do something like `api.get_status(status_id).retweeted_status.text`.

An additional dependency for a single utility method that re-implements existing functionality is excessive. Not only that, but the additions in this PR are outside the scope of this library and don't fit within its spirit or theme. The purpose of Tweepy is to provide an interface to Twitter's API, not methods to scrape the site itself.

Issue #829, opened 2017-01-02
StreamListener.filter(): `follow` argument doesn't accept integers for user ids


dannguyen:
This is for tweepy 3.5.0, Python 3.5.2

In regards to what's shown in this example: https://github.com/tweepy/tweepy/pull/827

      myStream.filter(follow=[2211149702])

That doesn't work in tweepy 3.5. Instead, you get this error message:

     TypeError: sequence item 0: expected str instance, int found

The problematic line is in tweepy/streaming.py:

              428         if follow:
       --> 429             self.body['follow'] = u','.join(follow).encode(encoding)

Expected behavior: the `follow` argument can handle integers or strings; in the former case, the `int` is stringified.

PR #830, opened 2017-01-03
Allow int or str IDs as streaming arguments

fitnr:
Fix for #829. The str.join method raises a `ValueError` for `int`s.


dannguyen:
Thanks @fitnr. I'm only just getting acquainted with using the Streaming API vs Tweepy so I know there are other dependencies and conventions to consider, so this is for future discussion: it might be worthwhile **to throw a warning** when a user passes in a string.

Right now, the behavior -- and I'm assuming this is unintended -- is to silently fail when a string value is passed into the `follow` argument/parameter. But because the `follow` parameter, according to Twitter's API spec, *only* accepts user IDs: 

https://dev.twitter.com/streaming/overview/request-parameters#follow

> A comma-separated list of user IDs, indicating the users whose Tweets should be delivered on the stream. 

In other words, it is never the case that the Twitter API will accept `follow=["ev"]`. Shouldn't the user be warned that what they are trying will *always* turn up nothing? 

In fact, I'd almost argue that it's better for `filter` to _throw an error_. Consider the uncommon case in which a user's **screen name** is entirely made up of numbers. Such as Twitter user [@101](https://twitter.com/101), whose __user_id__ is actually [3781411](http://gettwitterid.com/?user_name=101&submit=GET+USER+ID).

What behavior should the user expect with the following invocation?

       mystream.filter(follow=['101'])


Forcing the API of `filter` to accept *only* integers provides at least an implication of what the `follow` parameter is limited to:

       mystream.filter(follow=[101])

fitnr:
I don't think changing signature of the `stream.filter` function to now throw warnings or errors when sequences of `str`s are passed is a friendly choice. Since a Twitter ID may, for good reason, be stored either as an `int` or a `str`, I say let the [function accept either](https://docs.python.org/3.5/glossary.html#term-duck-typing), and throw an error if Twitter complains. Designing guard rails or trying to police a certain kind of behvaior seems unnecessary in a library that  tracks a complicated and ever-changing API.

To answer your question about this:
```python
mystream.filter(follow=['101'])
````
I would expect that will produce a stream that follows the user with `user_id == 101`. I would base that on reading the [API spec on streaming](https://dev.twitter.com/streaming/overview/request-parameters#follow). It should also be documented in the Tweepy docs (which are currently a little garbled on the streaming API, but that's another issue).

File: tweepy/streaming.py [hunk 1]
```diff
@@ -430,10 +430,13 @@ def filter(self, follow=None, track=None, async=False, locations=None,
         if self.running:
             raise TweepError('Stream object already connected!')
         self.url = '/%s/statuses/filter.json' % STREAM_VERSION
+
         if follow:
-            self.body['follow'] = u','.join(follow).encode(encoding)
+            self.body['follow'] = u','.join(str(f) for f in follow).encode(encoding)
```
Harmon758, line 435:
```suggestion
            self.body['follow'] = u','.join(map(str, follow)).encode(encoding)
```

tweepy/streaming.py
```diff
@@ -430,10 +430,13 @@ def filter(self, follow=None, track=None, async=False, locations=None,
...
+
         if track:
-            self.body['track'] = u','.join(track).encode(encoding)
+            self.body['track'] = u','.join(str(f) for f in track).encode(encoding)
```
Harmon758, line 438:
```suggestion
            self.body['track'] = u','.join(map(str, track)).encode(encoding)
```

Harmon758 (changes requested):

Harmon758 (approved):

PR #832, opened 2017-01-03
Use built in json library

fitnr:
Python has had a built-in json library since version 2.6, released in 2008. Tweepy's tests no longer include Python 2.5, so they're no reason to use simplejson.

This also removes a complicated import path that made every instance of Tweepy attempt to import a module that's not in requirements.

This commit also removes the json library as an attribute of the `Parser` object. Don't know what it was doing there, besides being unpythonic.

### DIFF
--- a/tweepy/parsers.py
+++ b/tweepy/parsers.py
@@ -4,8 +4,8 @@
 
 from __future__ import print_function
 
+import json as json_lib
 from tweepy.models import ModelFactory
-from tweepy.utils import import_simplejson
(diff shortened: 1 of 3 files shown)


ikalnytskyi:
If I'm not mistaken, `json` module bundled in Python2.6 is written in pure Python and hence pretty slow. There's no performance difference between `simplejson` and `json` starting with Python 2.7 and above though, and taking into account that Python 2.6 is dead – I support this change. :)

PR #849, opened 2017-02-23
Fills in testretweet() method

Peteous:
Tried to pull in the most recent tweet from the @twitter account, and then tries to retweet the tweet. If it fails, it skips the test. 

I understand that this isn't the requested *truly random* tweet, but in order to do that, you would need to test random 64-bit numbers until one of them was a valid tweet, then try to retweet it.

I wish I could find a legitimate tweet without using an outside method, but I don't think it's possible. I think this is a reasonably valid unit test.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -73,9 +73,9 @@ class TweepyAPITests(TweepyTestCase):
     def testretweetsofme(self):
         self.api.retweets_of_me()
 
+    @tape.use_cassette('testretweet.json')
     def testretweet(self):
-        # TODO(josh): Need a way to get random tweets to retweet.
-        raise SkipTest()
+        self.api.retweet(test_tweet_id)
 
     @tape.use_cassette('testretweets.json')
     def testretweets(self):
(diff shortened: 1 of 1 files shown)


Peteous:
There aren't any conflicts with the base branch, so this should be good to merge

tests/test_api.py
```diff
@@ -73,9 +73,13 @@ def testmentionstimeline(self):
     def testretweetsofme(self):
         self.api.retweets_of_me()
 
+    # @tape.use_cassette('testretweet.json')
     def testretweet(self):
-        # TODO(josh): Need a way to get random tweets to retweet.
-        raise SkipTest()
+    	try:
+    		tweet = self.api.user_timeline('twitter',count=1) # Get the latest tweet from @twitter
```
Harmon758:
```suggestion
    		tweet = self.api.user_timeline('twitter', count=1)  # Get the latest tweet from @Twitter
```

tests/test_api.py
```diff
@@ -73,9 +73,13 @@ def testmentionstimeline(self):
...
+    		self.api.retweet(tweet.id)
+    	except:
```
Harmon758: This blanket exception handler renders the test useless, since if any exceptions occur, the test is skipped.

Issue #865, opened 2017-03-21
set 'filter_level' for Streaming API has an error


sujcho:
Set *filter_level* for Streaming API as below gives an error:
```python
my_stream.filter(follow= follow_users,filter_level=['medium'],async=True)
```
Error says:
```bash
AttributeError: 'list' object has no attribute 'encode'
```
Python version: 3.5
Tweepy version: 3.6 (It also happens with 3.5 as well)

sujcho:
I looked at Streaming.py to see what gives an error:
```python
 if filter_level:
            self.body['filter_level'] = filter_level.encode(encoding)
```

**I think it has to be fixed as below, and it works. I tested it.**
```python
if filter_level:
            self.body['filter_level'] = u','.join(filter_level).encode(encoding)
```

sujcho:
Sorry, This was a mistake.

PR #866, opened 2017-03-21
fixed an error with filter_level

sujcho:
This belongs to issue #865
Setting filter_level for Streaming API as below gives an error:
AttributeError: 'list' object has no attribute 'encode'

Fixed to:
self.body['filter_level'] = u','.join(filter_level).encode(encoding)

and tested with Python version: 3.5


sujcho:
Sorry, the issue is closed.

Issue #883, opened 2017-05-12
tweepy errror filtering by coordinates


ptcmarin:
Traceback (most recent call last):
  File "analisis3.py", line 50, in <module>
    twitterStream.filter(locations=bogota)
  File "/usr/local/lib/python3.6/site-packages/tweepy-3.6.0-py3.6.egg/tweepy/streaming.py", line 450, in filter
  File "/usr/local/lib/python3.6/site-packages/tweepy-3.6.0-py3.6.egg/tweepy/streaming.py", line 364, in _start
  File "/usr/local/lib/python3.6/site-packages/tweepy-3.6.0-py3.6.egg/tweepy/streaming.py", line 297, in _run
  File "/usr/local/lib/python3.6/site-packages/six-1.10.0-py3.6.egg/six.py", line 686, in reraise
    raise value
  File "/usr/local/lib/python3.6/site-packages/tweepy-3.6.0-py3.6.egg/tweepy/streaming.py", line 266, in _run
  File "/usr/local/lib/python3.6/site-packages/tweepy-3.6.0-py3.6.egg/tweepy/streaming.py", line 323, in _read_loop
tweepy.error.TweepError: Expecting length, unexpected value found

ANY SUGGESTION?

wodim:
Paste your code?

Harmon758:
Closing in favor of #892 

For code block usage, see https://help.github.com/articles/creating-and-highlighting-code-blocks/.

Issue #876, opened 2017-04-13
Questions and multiple Error


shassanin:
Hi guys!

My actual code of tweepy is:

```
from tweepy.streaming import StreamListener

from tweepy import OAuthHandler

from tweepy import Stream

import json

access_token = "CENSORED"

access_token_secret = "CENSORED"

consumer_key = "CENSORED"

consumer_secret = "CENSORED"



class StdOutListener(StreamListener):
    def on_data(self, ecbdata):



        tweet = json.loads(ecbdata)

        created_at = tweet["created_at"]

        id_str = tweet["id_str"]

        text = tweet["text"]

        coordinates = tweet["coordinates"]

        obj = {"created_at": created_at, "id_str": id_str, "text": text,"coordinates": coordinates }

        tweetind = collection.insert_one(obj).inserted_id

        print (obj)

        return True

        def on_error(self, status):
            print (status)


if __name__ == '__main__':



    l = StdOutListener()

    auth = OAuthHandler(consumer_key, consumer_secret)

    auth.set_access_token(access_token, access_token_secret)

    stream = Stream(auth, l)


  

    from pymongo import MongoClient

    client = MongoClient()

    client = MongoClient('localhost', 27017)

    db = client['twitter_ecb']

    collection = db['twitter_ecb']


stream.filter(track=['ecb'])
```

I collect the data in MongoDB. For programing with Python i use Python3 in combination with the IDE PyCharm by JetBrains.

When i run this code, python console collects just one tweet but then it stops with following messages:

> Traceback (most recent call last):
>   File "C:/Users/Salah/PycharmProjects/TwitterData/tweets.py", line 76, in <module>
>     stream.filter(track=['ECB'])
>   File "C:\Users\Salah\AppData\Local\Programs\Python\Python36\lib\site-packages\tweepy-3.6.0-py3.6.egg\tweepy\streaming.py", line 450, in filter
>   File "C:\Users\Salah\AppData\Local\Programs\Python\Python36\lib\site-packages\tweepy-3.6.0-py3.6.egg\tweepy\streaming.py", line 364, in _start
>   File "C:\Users\Salah\AppData\Local\Programs\Python\Python36\lib\site-packages\tweepy-3.6.0-py3.6.egg\tweepy\streaming.py", line 297, in _run
>   File "C:\Users\Salah\AppData\Local\Programs\Python\Python36\lib\site-packages\six.py", line 686, in reraise
>     raise value
[...]

PR #885, opened 2017-05-16
Fix errors on keep-alive

DeanF:
Possibly the same issue as #883 #876


ikalnytskyi:
Hey! I wonder, how exactly does it help? Have you managed to catch the issue? What was that? A string with some whitespace characters instead of empty line? If so, why that was happened?

PR #895, opened 2017-06-11
Tidied up Streaming.py

Kevin-Gillanders:
Gave variables more descriptive names. made file PEP8 compliant

### DIFF
--- a/examples/streaming.py
+++ b/examples/streaming.py
@@ -3,33 +3,42 @@ from __future__ import absolute_import, print_function
 from tweepy.streaming import StreamListener
 from tweepy import OAuthHandler
 from tweepy import Stream
+from tweepy import API
 
-# Go to http://apps.twitter.com and create an app.
-# The consumer key and secret will be generated for you after
-consumer_key=""
-consumer_secret=""
 
-# After the step above, you will be redirected to your app's page.
-# Create an access token under the the "Your access token" section
-access_token=""
-access_token_secret=""
-
-class StdOutListener(StreamListener):
-    """ A listener handles tweets that are received from the stream.
-    This is a basic listener that just prints received tweets to stdout.
-
-    """
-    def on_data(self, data):
-        print(data)
-        return True
+# Override tweepy.StreamListener to add logic
+# both default methods on_status and on_error 
+# are overridden with added logic
+class MyStreamListener(StreamListener):
+    # A listener handles tweets that are received from the stream.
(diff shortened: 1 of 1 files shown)


examples/streaming.py
```diff
@@ -3,33 +3,42 @@
...
+auth.set_access_token(access_token, access_token_secret)
+api = API(auth)
 
 if __name__ == '__main__':
-    l = StdOutListener()
-    auth = OAuthHandler(consumer_key, consumer_secret)
-    auth.set_access_token(access_token, access_token_secret)
+    # Makes a MyStreamListener instance
+    myStreamListener = MyStreamListener()
+    myStream = Stream(auth = api.auth, listener = myStreamListener)
```
wjt: If you're aiming to be PEP 8-compliant, there should not be spaces around the `=`s for keyword parameters. From the [Other Recommendations](https://www.python.org/dev/peps/pep-0008/#other-recommendations) section:

> Don't use spaces around the = sign when used to indicate a keyword argument or a default parameter value.

fitnr: Also see [the naming conventions section](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments) for instances:

> lowercase with words separated by underscores as necessary to improve readability.

Harmon758, line 40: As noted:
```suggestion
    stream = Stream(auth=api.auth, listener=stream_listener)
```

examples/streaming.py
```diff
@@ -3,33 +3,42 @@
...
 
-    stream = Stream(auth, l)
-    stream.filter(track=['basketball'])
+    # In this example we will use filter to stream all tweets containing the word
+    # python. The track parameter is an array of search terms to stream.
```
Harmon758, line 43:
```suggestion
    # "python". The track parameter is a list of search terms to stream.
```

examples/streaming.py
```diff
@@ -3,33 +3,42 @@
...
+    myStream.filter(track = ['python'])
```
Harmon758, line 44:
```suggestion
    stream.filter(track=['python'])
```

Harmon758 (changes requested):
Using `API` isn't necessary for streaming and doesn't need to be added here.

Harmon758:
This is stale now, with https://github.com/tweepy/tweepy/commit/e3e1d062377a5833b16e8ac8976a56ac57fac8a2.

PR #896, opened 2017-06-15
Handles blank keep-alive new lines

jmartone:
Added .strip() to read_line() to handle keep-alive new lines. This is in line with the pip release of Tweepy here: https://pypi.python.org/pypi/tweepy/3.5.0

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -313,7 +313,7 @@ class Stream(object):
         while self.running and not resp.raw.closed:
             length = 0
             while not resp.raw.closed:
-                line = buf.read_line()
+                line = buf.read_line().strip()
                 if not line:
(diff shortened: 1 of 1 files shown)


ikalnytskyi:
@jmartone it's not safe to call `.strip()` on what's returned by `read_line()` since sometimes it's `None` and there were a fix to prevent crashes. Check out commit - https://github.com/tweepy/tweepy/commit/bb3d2075446b90fbb842266f68b0df91b623d09e.

ikalnytskyi:
Oops, sorry. It won't `None` anymore since commit https://github.com/tweepy/tweepy/commit/c1eddf1a5788ec4947d3a363dfe165ea599c31fc.

ikalnytskyi:
@Aaron1011 can you merge this patch please? it's very hard to use the lib without it.

PR #921, opened 2017-08-25
Add documentation for API.lookup_users

### LINKED ISSUE #539: Document API.lookup_users
It appears that the client supports the /users/lookup method, but it's just not referenced in the documentation. Is that intentional?

freeboson:
Resolves #539


Harmon758:
Resolves #539

Issue #840, opened 2017-02-03
tweet_mode=extended for statuses_lookup


vchrisb:
I used ```statuses_lookup``` to get specific tweets. 
But with [a recent change](https://dev.twitter.com/overview/api/upcoming-changes-to-tweets) of Twitter to have also ```extended tweets``` does the REST API only return a compacted version for these tweets.

The Twitter API now supports a parameter ```tweet_mode=extended``` to return not only the compacted version but also the extended one.

#785 does reference that it is currently not possible to use this parameter with ```statuses_lookup```

What needs to be changed to add ```tweet_mode=extended``` support to ```statuses_lookup```?

walmsley:
I've submitted a PR to resolve this: #926 - not sure who's maintaining this repo, but hope the change gets accepted soon!

Dmon1Unlimited:
please let this be fixed <3

ZeerakW:
I think @Aaron1011 is maintaining but don't know.

BigMistake:
I also have this problem. Is there any chance of an update on this?

shaharyar2643:
I'm also having the same problem. Is there a way around to have extended mode in api.statuses_lookcup

mjs2020:
Until this PR is reviewed/approved if you need to use @walmsley 's branch you can install it in your project with: `pip install git+https://github.com/walmsley/tweepy.git@patch-1` (kudos @walmsley  for the work, it saved me a lot of time today)

lwahedi:
@walmsley 's branch doesn't install with the latest pip. I just created a pull request to change the setup.py file so that it should work, as per here: 
https://stackoverflow.com/questions/25192794/no-module-named-pip-req

lwahedi:
@jalperin In the terminal/Bash, first pip uninstall tweepy, then run the above in terminal. Otherwise you'll have two tweepy installs and the original will probably get imported instead when you try to use it.

lwahedi:
Now at https://github.com/lwahedi/tweepy.git@master

jason-lange:
Hi, I tried to install this package using what I understood of lwahedi's guidance. I uninstalled tweepy and entered in terminal: 
    pip install git+https://github.com/lwahedi/tweepy.git@master
But this gets me an error. 
     xcrun: error: invalid active developer path (/Library/Developer/CommandLineTools), missing xcrun 
     at: /Library/Developer/CommandLineTools/usr/bin/xcrun
[...]

PR #926, opened 2017-09-11
Allow tweet_mode=extended in statuses_lookup

walmsley:
In response to #840


Dmon1Unlimited:
please let this be fixed <3
I have a large batch of tweets to pull which i assume.
statuses_lookup takes some time but if extended mode isn't ready i'd have to change my code and get it done one at a time via: api.get_status(798508477572222976,tweet_mode='extended') which is painstakingly slow

PR #939, opened 2017-10-05
Added possibility to compare `eq` and `ne` between users, using the unique `id`s.

gardir:
The code is copied from `Status`, with only the _type_ changed in `instanceof` from `Status` to `User`.

### DIFF
--- a/tweepy/models.py
+++ b/tweepy/models.py
@@ -201,6 +201,20 @@ class User(Model):
                                        *args,
                                        **kargs)
 
+    def __eq__(self, other):
+        if isinstance(other, User):
+            return self.id == other.id
+
+        return NotImplemented
+
+    def __ne__(self, other):
+        result = self == other
+
+        if result is NotImplemented:
+            return result
+
+        return not result
+
 
 class DirectMessage(Model):
 


fitnr:
This is a good idea. As written, the `NotImplemented` will be returned whenever the `other` is a different type of object. Shouldn't there be a check that other == self, to allow for the possibility that the other object has an `__eq__` method that understands `Status` objects?

gardir:
I am not sure if I understand you correctly @fitnr -- are you thinking about the situation:
`other == self  # true`, but `isinstance(other, User)  # false`  ?
I am not sure such a situation could occur _within_ the `User.__eq__` scope.
Then again, I may have misunderstood your comment or just not be that good of a Python programmer.
Please ellaborate or provide `code` if so.

fitnr:
Nevermind, I didn't know a detail about how `__eq__` works: if `NotImplemented` is returned, Python automatically checks for `other.__eq__`.

This looks great, my only suggestion would be to add comparable functions on the `Status` object.

gardir:
Is this new commit (2f8766162563415cf3f6bf6b398050c176768b02) the comparable functions you wanted on the `Status` object? Or are there some comparability between `User` and `Status` object I am not aware of you are refering too?

fitnr:
No, I was talking about Status-to-Status functions. 👍

Harmon758:
I don't think comparison operations between Status objects are necessary or make much sense, so I've gone ahead and reverted that commit.

Issue #958, opened 2017-10-27
cursor .pages() with api.search_users returning same page again and again


manji369:
```
auth = tweepy.OAuthHandler(consumer_token, consumer_secret)
auth.set_access_token(access_token, access_secret)
api = tweepy.API(auth)
user_objs = []
name = "jacqueline farkas"
id_strs = {}
page_no = 0
try:
    for page in tweepy.Cursor(api.search_users, name).pages(3):
        dup_count = 0
        print("*******  Page", str(page_no))
        print("Length of page", len(page))
        user_objs.extend(page)
        for user_obj in page:
            id_str = user_obj._json['id_str']
            if id_str in id_strs:
                # print("Duplicate for:", id_str, "from page number:", id_strs[id_str])
                dup_count += 1
            else:
                # print(id_str)
                id_strs[id_str] = page_no
        time.sleep(1)
        print("Duplicates in page", str(page_no), str(dup_count))
        page_no += 1
except Exception as ex:
    print(ex)```

With the above code, I am trying to get the search results for users using tweepy(Python 3.5.2, tweepy 3.5.0) cursor. The results are being duplicated with the pages parameter being passed. Is it the right way to query the search_users using the tweepy cursor? I am getting results for the above code with the following pattern:

`1. for low search results(name = "jacqueline farkas") (There are actually 9 results returned for manual search on twitter website):

    *******  Page 0
    Length of page 2
    Duplicates in page 0 0
    *******  Page 1
    Length of page 2
    Duplicates in page 1 2
    *******  Page 2
    Length of page 2
    Duplicates in page 2 2
    *******  Page 3
    Length of page 2
    Duplicates in page 3 2

2. for high search results (name = "jon snow")

    *******  Page 0
    Length of page 20
    Duplicates in page 0 0
    *******  Page 1
    Length of page 20
    Duplicates in page 1 20
    *******  Page 2
    Length of page 20
    Duplicates in page 2 0
    *******  Page 3
    Length of page 20
    Duplicates in page 3 0

3. With no limit on pages specified (.pages()), the same page is being repeated and never stops. Do I have to change the page somewhere everytime i loop over the cursor?`

manji369:
[...]

PR #959, opened 2017-10-28
PageIterator: Change initial page number to 1 and handled duplicates

manji369:
Initialized self.current_page to 1 instead of 0 and handled duplicates from previous pages to return unique results and to determine the stop iteration criteria (Refer issue [958](https://github.com/tweepy/tweepy/issues/958)).

### DIFF
--- a/tweepy/cursor.py
+++ b/tweepy/cursor.py
@@ -159,16 +159,38 @@ class PageIterator(BaseIterator):
 
     def __init__(self, method, args, kargs):
         BaseIterator.__init__(self, method, args, kargs)
-        self.current_page = 0
+        self.current_page = 1
+        self.prev_page_ids = set()
+        self.current_page_ids = set()
+        self.flag = False
 
     def next(self):
+        if self.flag:
+            raise StopIteration
+
+        # Setting previous page and initailizing current page
+        self.prev_page_ids, self.current_page_ids = self.current_page_ids, set()
+
         if self.limit > 0:
             if self.current_page > self.limit:
                 raise StopIteration
 
         items = self.method(page=self.current_page, *self.args, **self.kargs)
+        # Removing duplicates in current page and setting flag to indicate
+        # end of results
+        temp_items = []
+        for item in items:
+            current_id = item._json['id']
+            if current_id in self.prev_page_ids:
+                self.flag = True
+            else:
+                self.current_page_ids.add(current_id)
+                temp_items.append(item)
+        items = temp_items
+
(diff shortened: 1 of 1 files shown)


TimotheeJeannin:
Is this going to me merged soon ?

jaredthecoder:
@Harmon758 I am going to test these changes locally with search_users() and I'll let you all know if it still works.

jaredthecoder:
Screen Shot 2019-10-18 at 11 26 53

Seems to be working.

meow232309:
Hi,
So i'm still seeing the same issue here and don't believe its fixed. Anyone else having issues?

thanks

meow232309:
> Screen Shot 2019-10-18 at 11 26 53
> 
> Seems to be working.

I just did the same thing with a different query for page 1 and 2 and got the exact same results.

Harmon758 (changes requested):
This should fix two separate bugs (#958). The first being simply the initial page number incorrectly being 0 instead of 1, and the second being [a longstanding issue with Twitter's API that won't be fixed before a Twitter API v2 equivalent endpoint is released](https://twittercommunity.com/t/odd-pagination-behavior-with-get-users-search/148502) (#1465).

I think the [GET users/search endpoint](https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/follow-search-get-users/api-reference/get-users-search) is the last remaining endpoint to still use the `page` parameter, so since the `PageIterator` will likely only be used for [`API.search_users`](https://tweepy.readthedocs.io/en/latest/api.html#API.search_users), there shouldn't be any performance impacts on using a `Cursor` with any other methods due to any implementation that fixes the latter bug, even if the GET users/search endpoint is the only one that exhibits it.

Since the latter bug causes identical pages after the last, removing duplicates and using a flag to raise `StopIteration` on the page after shouldn't be necessary, as it would just return an extra empty page. `StopIteration` can just be raised immediately when duplicates are encountered. This should also remove the need for a `temp_items`.

Using two sets of IDs also shouldn't be necessary, as the current list of items/users can simply be compared with the previous one.

There are also conflicts that need to be resolved now.

Harmon758:
@meow232309 Are you using the version in this PR?

Harmon758:
This should now be resolved by https://github.com/tweepy/tweepy/commit/a7d1608903825a4c2d781e5f8f709f26ed1bad03 and https://github.com/tweepy/tweepy/commit/3c1a397cd8d5bb81ff245ca0645c6462e57a5d33.

PR #987, opened 2018-01-10
Fix typo in example.

skion:


skion:
@joshthecoder Looking at it now I don't think this change actually made sense

joshthecoder:
Ah you right now that I read the comment above more closely. I reverted the change in master. No problem!

PR #1019, opened 2018-03-28
adding page parameter to search function

mattalytics:
Updating search function to match tweepy documentation


joshthecoder:
Doesn't the endpoint use max_id and since_id to page through results? I don't see `page` documented for that endpoint.

Harmon758:
`page` is an outdated parameter that was removed with https://github.com/tweepy/tweepy/commit/2098078d8634012c92e9f00a0fdfd9ea07b2236b in Tweepy 2.1.
It is no longer a valid parameter for the [standard search API](https://developer.twitter.com/en/docs/tweets/search/api-reference/get-search-tweets).
Instead, the documentation for the parameter should be removed. See #1199.

Issue #1029, opened 2018-04-14
pip 10 build failure


FRidh:
This package (latest stable, master) fails to build with pip 10 because it imports something which is now internal.

`setup.py`
```
from pip.req import parse_requirements
```

Harmon758:
This should be fixed by https://github.com/tweepy/tweepy/commit/29e32cdf4a67b1fca25dbf04d7b8069ac1c9dd0d (#1030) in [3.7.0](https://github.com/tweepy/tweepy/releases/tag/3.7.0).

PR #1030, opened 2018-04-18
#1029 - migrate requirements to setup.py, reference in requirements.txt

bcipolli:
`parse_requirements` is no longer available from `pip`. Followed guidance from https://caremad.io/posts/2013/07/setup-vs-requirement/ to work around this / fix the design bug in putting dependencies in `requirements.txt` instead of `setup.py`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-requests>=2.11.1
-requests_oauthlib>=0.7.0
-six>=1.10.0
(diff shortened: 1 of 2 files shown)


joshthecoder:
Looks good to me. I don't work in Python much so I'm a bit out of date with how packaging works. The method described in that post seems correct. 👍

PR #1036, opened 2018-04-26
updated cursor_tutorial with extended_tweet instructions

TimmDay:
Tweepy, thank you for your work.

I ran into a tweet-truncation hurdle while collecting tweets for a Hungarian language corpus.
tweets greater than 140 characters are truncated back down to 140 characters in the default use case.

This will be a very common problem for computational linguists, and is very easy to overlook until it is too late (ie your corpus is being used for sentiment analysis or some sort of processing down the line).

I could not find proper documentation on how to fix the issue, so I thought I would propose an update to the docs. Hopefully it saves some students some stress and some linguists some lost time.

Thank you again for your work,
Tim


joshthecoder:
Thanks for the contribution!

PR #1037, opened 2018-04-27
Update count parameter for search_users and list_timeline

data4d:
Twitter's API no longer supports "per_page" as a parameter for search_users and list_timeline. Instead, "count" must be used. This is reflected in the most recent Twitter API Documentation. I validated that the "count" parameter works for both search_users and list_timeline using Tweepy 3.6.


joshthecoder:
Nice! Thanks for the contribution

Issue #1017, opened 2018-03-17
async invalid syntax


erooskie:
I'm fairly new to coding---just taking a coding class this semester. I'm using a MacBook Air and downloaded PyCharm, and so far, I've been able to figure most things out, but I'm stuck on this one.

I'm trying to get the tweepy Stream Listener to return the tweets for a certain hashtag, and this code worked without issue on the Windows computers on campus. 
The code I'm running looks like this:

    import tweepy
    #override tweepy.StreamListener to add logic to on_status
    class MyStreamListener(tweepy.StreamListener):

        def on_status(self, status):
            print(status.text)
    
        def on_error(self, status_code):
            if status_code == 420:
                # returning False in on_data disconnects the stream
                return False
    auth = tweepy.OAuthHandler(consumer_key, consumer_secret)
    auth.set_access_token(access_token, access_token_secret)
    
    myStreamListener = MyStreamListener()
    myStream = tweepy.Stream(auth, listener=myStreamListener)
    
    myStream.filter(track=['#fakenews'])

On my computer, it returns this error:

> Traceback (most recent call last):
>  File "/Users/USERNAME/PycharmProjects/Homework/Twitter.py", line 7, in <module>
    import tweepy
  File "/Users/USERNAME/PycharmProjects/Homework/venv/lib/python3.7/site-packages/tweepy/__init__.py", line 17, in <module>
    from tweepy.streaming import Stream, StreamListener
  File "/Users/USERNAME/PycharmProjects/Homework/venv/lib/python3.7/site-packages/tweepy/streaming.py", line 358
    def _start(self, async):
                                     ^
SyntaxError: invalid syntax

The carat is actually under `async`, not at the beginning of the line. The last two "files" are the code from the tweepy module, not any code I've written or searched through.

I've tried changing the interpreter for my project in PyCharm, but I'm not sure what else to try. My professor looked at it and didn't want to get into the backend code (and honestly, I don't either, since I don't really know what I'm doing), so I'm stuck as to what to try next or actually what the real issue is. 

Any suggestions?

stoictrader:
The same issue but on Windows10 x64 python3.7 beta, on python3.6 it works fine, so seems like incompatibility with python 3.7?

hugovk:
https://github.com/tweepy/tweepy/pull/1042 has been merged. Can this be closed?

skidzo:
[...]

PR #1042, opened 2018-05-12
Fix Python3.7 compatibility issue

nyanye:
passing "async" argument can make invalid argument error on python 3.7
#1017 async invalid syntax #1017

### DIFF
--- a/tests/test_streaming.py
+++ b/tests/test_streaming.py
@@ -105,7 +105,7 @@ class TweepyStreamTests(unittest.TestCase):
 
     def test_track_encoding(self):
         s = Stream(None, None)
-        s._start = lambda async: None
+        s._start = lambda is_async: None
         s.filter(track=[u'Caf\xe9'])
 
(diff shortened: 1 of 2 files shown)


joshthecoder:
Looks good thanks :)

alexott:
Can we get a release with this bug fix?

shuuji3:
For those who need to use Tweepy in Python3.7 immediately, here is a workaround:
```bash
$ pip install git+https://github.com/tweepy/tweepy
```
This command will install the GitHub's latest version (as v3.6.0) instead of PyPI's v3.6.0.

ilf:
Python 3.7 hat been out for more than three months now, so have the initial patches here. Is there anything we can do to help get out a tweepy release that works with current Python? Thanks!

denyramandaa:
Thank you! it's helped me!

hugovk:
There is now a Tweepy release supporting Python 3.7.

Thank you, @joshthecoder!

Issue #810, opened 2016-11-18
application raises Too many open files (errno=24)


sam-s:
`lsof` shows a gazillion open sockets:
```
python  17055  sds  286u  IPv4 272626       0t0     TCP datasci-1:48948->104.244.42.194:https (ESTABLISHED)
python  17055  sds  287u  IPv4 272361       0t0     TCP datasci-1:45474->104.244.42.130:https (ESTABLISHED)
python  17055  sds  288u  IPv4 272373       0t0     TCP datasci-1:48956->104.244.42.194:https (ESTABLISHED)
python  17055  sds  289u  IPv4 272369       0t0     TCP datasci-1:44114->104.244.42.2:https (ESTABLISHED)
python  17055  sds  290u  IPv4 272377       0t0     TCP datasci-1:45482->104.244.42.130:https (ESTABLISHED)
python  17055  sds  291u  IPv4 272381       0t0     TCP datasci-1:52306->104.244.42.66:https (ESTABLISHED)
python  17055  sds  292u  IPv4 273409       0t0     TCP datasci-1:44122->104.244.42.2:https (ESTABLISHED)
python  17055  sds  293u  IPv4 273413       0t0     TCP datasci-1:48964->104.244.42.194:https (ESTABLISHED)
python  17055  sds  294u  IPv4 273417       0t0     TCP datasci-1:45490->104.244.42.130:https (ESTABLISHED)
python  17055  sds  295u  IPv4 273421       0t0     TCP datasci-1:52314->104.244.42.66:https (ESTABLISHED)
python  17055  sds  296u  IPv4 272630       0t0     TCP datasci-1:44130->104.244.42.2:https (ESTABLISHED)
python  17055  sds  297u  IPv4 272634       0t0     TCP datasci-1:48972->104.244.42.194:https (ESTABLISHED)
python  17055  sds  298u  IPv4 272638       0t0     TCP datasci-1:45498->104.244.42.130:https (ESTABLISHED)
python  17055  sds  299u  IPv4 272642       0t0     TCP datasci-1:52322->104.244.42.66:https (ESTABLISHED)
python  17055  sds  300u  IPv4 272646       0t0     TCP datasci-1:44138->104.244.42.2:https (ESTABLISHED)
python  17055  sds  301u  IPv4 272650       0t0     TCP datasci-1:48980->104.244.42.194:https (ESTABLISHED)
[...]

PR #1068, opened 2018-07-05
Wrap APIMethod.execute in try/finally to ensure that the session is closed even on error.

sam-s:
This fixes https://github.com/tweepy/tweepy/issues/810.

### DIFF
--- a/tweepy/binder.py
+++ b/tweepy/binder.py
@@ -124,7 +124,7 @@ def bind_api(**config):
 
                 self.path = self.path.replace(variable, value)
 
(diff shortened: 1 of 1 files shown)


Harmon758:
Thanks for the PR!
I've done this in a different way with https://github.com/tweepy/tweepy/commit/3560a52aef9415d8d6297130d2495fd99e1bf291, so as to also close the session when using a Cursor, specifically an IdIterator, where the method is called with `create=True` and `execute` is never called.

Issue #697, opened 2016-01-28
Update API.is_list_member -> API.show_list_member in documentation


johngriffin:
Hi,

I'm observing the following with v 3.5.0:

`api.list_members(settings['screen_name'], settings['list_slug'])`
This returns as per docs.

`api.is_list_member(settings['screen_name'], settings['list_slug'], 2916305152)`
Returns: `AttributeError: 'API' object has no attribute 'is_list_member`

There is no "is_list_member" in tweepy/api.py, however there is a show_list_member:

```
show_list_member(self):
        """ :reference: https://dev.twitter.com/docs/api/1.1/get/lists/members/show
            :allowed_param:'list_id', 'slug', 'user_id', 'screen_name',
             'owner_screen_name', 'owner_id
```

Additionally it seems that there is an issue with get_list, suggesting that there is some broader issue with the list methods:

`api.get_list(screen_name=settings['screen_name'], slug=settings['list_slug'])`

returns "tweepy.error.TweepError: [{u'message': u'You must specify either a list ID or a slug and owner.', u'code': 112}]"

awlamb:
You're correct - get_list is broken. It works fine if you pass through a list id but breaks when you pass through owner_screen_name and slug params

mb3152:
what is the best way to get the list id if I just have the name and owner of the list?

ramanadan36:
Please check
https://stackoverflow.com/questions/19213285/tweepy-2-1-add-member-to-a-list

Harmon758:
`API.is_list_member` was replaced by `API.show_list_member` with https://github.com/tweepy/tweepy/commit/a5f75d89a820242e439d315b86a2c2d849505b9f in Tweepy 2.0 to reflect changes in Twitter's API. The documentation is just out of date.

As for [`API.get_list`](https://tweepy.readthedocs.io/en/latest/api.html#API.get_list), the parameter name is `owner_screen_name`, not `screen_name`:
https://github.com/tweepy/tweepy/blob/1de28d24ec8fce97c3775c0795c97d70ca6d3a5c/tweepy/api.py#L1028-L1038  See https://developer.twitter.com/en/docs/accounts-and-users/create-manage-lists/api-reference/get-lists-show.

For updating the documentation for API.get_list, see #1201.

@awlamb I'm unable to reproduce this. Can you provide a reproducible example?
Something like `api.get_list(owner_screen_name="Twitter", slug="developers")` works for me.

[...]

PR #1077, opened 2018-07-16
Add is_list_member method

jbaldo:
Implementation a bit sloppy, but a method like this would provide a function that matches the [is_list_function described in the docs](http://docs.tweepy.org/en/v3.6.0/api.html#API.is_list_member) / resolve Issue #697.

Twitter replies to `show_list_member` with this error when the user isn't in the list:
```python
tweepy.error.TweepError: [{'code': 109, 'message': 'The specified user is not a member of this list.'}]
```

### DIFF
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -1148,6 +1148,17 @@ class API(object):
                            'owner_screen_name', 'owner_id']
         )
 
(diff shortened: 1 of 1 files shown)


Harmon758:
This seems like an unnecessary convenience function for something the user should handle.
Instead, the documentation should simply be updated to reflect the removal of `API.is_list_member`.
See #697.

PR #1089, opened 2018-08-10
Giving options for the `__repr__` of  `Model` class

anbarief:
For convenience in viewing the object representation.

Example : let `x` be a `tweepy.models.Status` object
```
>>> x.object_repr
False
>>> x
Status(in_reply_to_status_id_str=None, in_reply_to_user_id=None, place=None, 
....
....
*******quite lengthy*******
....
....
 'lang': 'en'})
```
If you want the short version, we can set the `object_repr=True`.

```
x.object_repr=True
>>> x
<tweepy.models.Status object at 0xb17562ac>
```

This will apply for all subclasses of `Model`.

Regards.

### DIFF
--- a/tweepy/models.py
+++ b/tweepy/models.py
@@ -38,6 +38,7 @@ class Model(object):
 
     def __init__(self, api=None):
         self._api = api
+        self.object_repr = False
 
     def __getstate__(self):
         # pickle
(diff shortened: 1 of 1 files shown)


Harmon758:
I'm not sure what the benefits of this PR are.
The current representation matches the standard of looking ["like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment)."](https://docs.python.org/3/reference/datamodel.html#object.__repr__)
The default representation is generally not very useful, and the changes in this PR seem like anti-patterns.

PR #1090, opened 2018-08-11
Add a support of the 'perform_block' param for api.report_spam() method

shuuji3:
Now tweepy's `api.report_spam()` method only accept `user_id` and `screen_name` params. So when we report spam for an account, the account always be blocked. This PR supports the `perform_block` param and enable to report spam without blocking the target account (by setting it to `False`).

See also: [POST users/report_spam — Twitter Developers](https://developer.twitter.com/en/docs/accounts-and-users/mute-block-report-users/api-reference/post-users-report_spam)


shuuji3:
@Harmon758 Thanks for merging!

Issue #773, opened 2016-08-08
Streaming does not gracefully terminate if no messages are received


jusw85:
Using tweepy==3.5.0

```
def signal_handler(signal, frame):
    print("Ctrl-C received")
    myStream.disconnect()

myStreamListener = MyStreamListener()
myStream = tweepy.Stream(auth=api.auth, listener=myStreamListener)

signal.signal(signal.SIGINT, signal_handler)
myStream.filter(track=['keywords'])
```

I would like to Ctrl-C terminate my application. Calling `stream.disconnect()` sets `running=False`. However, this variable is only tested when Tweepy receives a message. If no messages are received, it blocks while reading from the stream (specifically `self._buffer += self._stream.read(self._chunk_size)` in `read_line()`), and I have to force kill my application.

Harmon758:
With https://github.com/tweepy/tweepy/commit/37e701e2193dcf617e7235d3b004f790c5ff784e, the stream should now be able to disconnect on the next keep-alive signal received.

PR #1105, opened 2018-09-19
Allow to close stream on keep-alive messages (Fixes #897 #773)

### LINKED ISSUE #897: Streaming: disconnect on keep_alive()
Hi

I have a use case where I want to connect to the streaming API for a period of time and then disconnect even when no tweet is received during that period.

When no tweet is received, I can check if defined duration has elapsed in StreamListener keep_alive(), but there is no mechanism to disconnect in this case since _read_loop() is not reading any return value from keep_alive() and there is no way to break the loop in this situation.

As a workaround I am rising a custom exception in my implementation of keep_alive() but this is not ideal, I understand that a better solution is to let keep_alive() return a boolean that when is false breaks the read loop, same as on_data(), on_status(), etc.

lbesuchet:
This fix allows to disconnect the stream at the next keep-alive message received.
This is possible by either calling disconnect() or having keep_alive() returning False.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -35,3 +35,4 @@ Kohei YOSHIDA
 Mark Smith (@judy2k)
 Steven Skoczen (@skoczen)
 Samuel (@obskyr)
+Luc Besuchet (@lbesuchet)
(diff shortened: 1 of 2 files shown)


Harmon758:
Thanks, but I've resolved this in a better way with https://github.com/tweepy/tweepy/commit/37e701e2193dcf617e7235d3b004f790c5ff784e, without reducing the default chunk size.
See also https://github.com/tweepy/tweepy/issues/897#issuecomment-765578161.

PR #1122, opened 2018-10-23
Makes tweepy compatible with python 3.7

simonRos:
'async' is now a reserved word/key word. I've renamed the async variables in streaming.py to asyncro in order to avoid syntax errors.


Harmon758:
#1042 already addressed this. `is_async` should be fine.
You're also introducing a syntax error with the last line.

joshthecoder:
Closing as it seems this isn't needed. If other 3.7 issues arise open an issue.

PR #1124, opened 2018-10-24
remove harcoded api url

osmay88:
This is hardcoded URL won't allow mocking the streaming API.


joshthecoder:
Thanks for the contribution!

PR #1126, opened 2018-11-07
Allow optionally setting daemon flag on async threads.

mmontagna:
Would you be open to a PR which sets the daemon flag on async threads?

https://docs.python.org/3/library/threading.html#threading.Thread.daemon

This will allow python scripts to exit cleanly when all non-daemon threads are stopped. The current behavior is that the streaming thread stays up and prevents the script from exiting.

[EDIT]
Changed PR to optionally allow setting daemon flag, but maintain the default behavior of setting it to false.

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -364,6 +364,7 @@ class Stream(object):
         self.running = True
         if is_async:
             self._thread = Thread(target=self._run)
+            self._thread.daemon = True
             self._thread.start()
         else:
             self._run()
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -195,6 +195,7 @@ class Stream(object):
         self.auth = auth
         self.listener = listener
         self.running = False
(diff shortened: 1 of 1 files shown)


mmontagna:
Ahh I see this PR already existed: https://github.com/tweepy/tweepy/pull/268

mmontagna:
Ahh I see this PR already existed: https://github.com/tweepy/tweepy/pull/268

mmontagna:
What if it was added as an option which could be passed to the functions which start async threads?

joshthecoder:
Yeah I think adding this as an option makes sense. How about we pass it to the "options" object the constructor receives? Then you can just assign it to `self` like we do for the other options (default is false).

mmontagna:
@joshthecoder Alright I've updated the PR to make it an option to the Stream constructor. What do you think now?

Also are there any docs I should update?

joshthecoder:
Looks good now thanks! Sorry for the long wait on seeing this again

PR #1145, opened 2018-12-23
Correction on is_async argument of filter function

osainz59:
The documentation shows how to use filter's is_async argument, but it calls as async and not as is_async. Just a minor issue.


osainz59:
Okey is duplicated, sorry.

PR #1152, opened 2019-01-14
Update streaming_how_to.rst

kenniy:
minor change in documentation: filter parameter should be `is_async` instead of `async`


joshthecoder:
Closing as this was fixed in another PR that was merged. Thanks for sending!

PR #1158, opened 2019-01-21
set access token through it's method maybe tedious

codingpy:
Because the only method that would influence the OAuthHandler instance's access token is set_access_token, adding a shortcut that set the access token in intantialization would be more user-friendly ? (as far as I'm concerned)
It is just a small modification, thanks for your patience on reading this comments...

### DIFF
--- a/tweepy/auth.py
+++ b/tweepy/auth.py
@@ -31,7 +31,7 @@ class OAuthHandler(AuthHandler):
     OAUTH_HOST = 'api.twitter.com'
     OAUTH_ROOT = '/oauth/'
 
-    def __init__(self, consumer_key, consumer_secret, callback=None):
+    def __init__(self, consumer_key, consumer_secret, callback=None, access_token=None, access_token_secret=None):
         if type(consumer_key) == six.text_type:
             consumer_key = consumer_key.encode('ascii')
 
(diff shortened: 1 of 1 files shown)


MichaelCurrin:
The change looks good to me. 

Though the fork repo has been deleted.

Also curious as to why the `.encode` is applied to consumer tokens and not access tokens even in the original flow (in `.set_access_token`). No maybe its no needed, or it should be added to `__init__` for the access tokens for consistency.

tweepy/auth.py
```diff
@@ -31,7 +31,7 @@ class OAuthHandler(AuthHandler):
     OAUTH_HOST = 'api.twitter.com'
     OAUTH_ROOT = '/oauth/'
 
-    def __init__(self, consumer_key, consumer_secret, callback=None):
+    def __init__(self, consumer_key, consumer_secret, callback=None, access_token=None, access_token_secret=None):
```
Harmon758, line 34:
```suggestion
    def __init__(self, consumer_key, consumer_secret, access_token=None,
                 access_token_secret=None, callback=None):
```

Harmon758 (changes requested):
There are conflicts that need to be resolved now.

Harmon758:
Superseded by https://github.com/tweepy/tweepy/commit/99f3583d99ac9a0003273318e7628235bba707f0

PR #1069, opened 2018-07-05
update doc: async --> is_async

sam-s:

PR #1160, opened 2019-01-22
filter method only accepts is_async, not async

orekiHoutarou:
The definition of the Stream class's filter method for now only accept is_async, which also means that the documentation should follow up.
Thanks for your great work, hope I could make a few of help.


Harmon758:
Duplicate of #1069

orekiHoutarou:
Thanks for your review!

PR #1138, opened 2018-12-04
Fix Cursor Tutorial documentation

Harmon758:

PR #1181, opened 2019-03-12
fix indentation and improve comment

Uinelj:
fix indentation of code block l.106-109
add a more meaningful comment in the code block


Harmon758:
Partially duplicate of #1138.
I don't think an explanatory comment is necessary since it's explained in the paragraph prior.
This specific comment could also be misleading, as it leaves out the change from `.text` to `.full_text`.

PR #1208, opened 2019-05-04
Add cursor param to blocks_ids and mutes_ids

okomarov:
Twitter has added cursor support to both [mutes_ids](https://developer.twitter.com/en/docs/accounts-and-users/mute-block-report-users/api-reference/get-mutes-users-ids#parameters) and [blocks_ids](https://developer.twitter.com/en/docs/accounts-and-users/mute-block-report-users/api-reference/get-blocks-ids#parameters).

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -477,11 +477,12 @@ Block Methods
    :rtype: list of :class:`User` objects
 
 
-.. method:: API.blocks_ids()
+.. method:: API.blocks_ids([cursor])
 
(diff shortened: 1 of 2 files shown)


okomarov:
@Harmon758 This is a very minimalistic change. We had to fork the library and add this change ourselves and we are using it successfully but would prefer it integrated upstream, as it would simplify our dependencies.

Harmon758:
Merged @okomarov

okomarov:
Great thanks @Harmon758

PR #1211, opened 2019-05-07
Fix help() calls on the instantiated API, by remove confusing propert…

johnkw:
…y decorators from some of the member functions, which are not actually property descriptors.

Consolidate function docs for legacy _lookup_users() _lookup_users() functions, which can now be removed given the way bind_api has been used since 2014, which also improves help() for them.

This also makes the nature of pagination more obvious and explicit.

### DIFF
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -78,8 +78,7 @@ class API(object):
                 )
             )
 
-    @property
-    def home_timeline(self):
+    def home_timeline(self, *args, **kwargs):
         """ :reference: https://developer.twitter.com/en/docs/tweets/timelines/api-reference/get-statuses-home_timeline
             :allowed_param:'since_id', 'max_id', 'count'
         """
@@ -89,15 +88,15 @@ class API(object):
             payload_type='status', payload_list=True,
             allowed_param=['since_id', 'max_id', 'count'],
             require_auth=True
-        )
+        )(*args, **kwargs)
+    home_timeline.pagination_mode = 'id'
 
     def statuses_lookup(self, id_, include_entities=None,
                         trim_user=None, map_=None, tweet_mode=None):
         return self._statuses_lookup(list_to_csv(id_), include_entities,
                                      trim_user, map_, tweet_mode)
 
-    @property
-    def _statuses_lookup(self):
+    def _statuses_lookup(self, *args, **kwargs):
         """ :reference: https://developer.twitter.com/en/docs/tweets/post-and-engage/api-reference/get-statuses-lookup
             :allowed_param:'id', 'include_entities', 'trim_user', 'map', 'tweet_mode'
         """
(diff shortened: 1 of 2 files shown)


okomarov:
It was unclear what you meant by improved `help()`:

BEFORE
```python
help(tweepy.api.home_timeline)

Help on function _call in module tweepy.binder:

_call(*args, **kwargs)
```

AFTER
```python
help(tweepy.api.home_timeline)

Help on method home_timeline in module tweepy.api:

home_timeline(*args, **kwargs) method of tweepy.api.API instance
    :reference: https://developer.twitter.com/en/docs/tweets/timelines/api-reference/get-statuses-home_timeline
    :allowed_param:'since_id', 'max_id', 'count'
```

This is nice, +1. 

However, you have the pagination settings in API and I think @joshthecoder (@Harmon758) intended to keep that file as close as possible to the twitter API, i.e. just as a simple mapping.

johnkw:
> However, you have the pagination settings in API and I think @joshthecoder (@Harmon758) intended to keep that file as close as possible to the twitter API, i.e. just as a simple mapping.

The paging style has always been stored in api.py, it's just that previously the "cursor" information was guessed at with some unusual logic in the binder.py file.  At least now it's more obvious what's going on, and the code is simpler.

I merged the fork so the pull request should be ready to commit again.

okomarov:
It is more obvious but this way the twitter api becomes polluted with tweepy internals. The Twitter API uses `max_id`, `since_id`, `cursor` and `page` while `pagination_mode` is tweepy only.

The correct way would be to keep the separation between `api.py` and tweepy internals. Remove pagination_mode altogether and let the tweepy Cursor do the heavy lifting/guessing with max_id`, `since_id`, `cursor` and `page`.

johnkw:
Yes, as I alluded to in issue #1212 it would be good to clean up api.py so that the function definition isn't the place that pagination_mode has to be stored. My plan was to create a patch that dynamically generates the API member functions in api.py from a dict().  This would get rid of all the boiler-plate bind_api repetition that makes api.py hard to read. The Cursor code would be able to access that same dict() to get the pagination style. In any case, one cleanup at a time. If the project is still active and accepting patches I'll keep going.

tweepy/api.py
```diff
@@ -334,27 +328,19 @@ def lookup_users(self, user_ids=None, screen_names=None, include_entities=None,
...
             api=self,
             path='/users/lookup.json',
             payload_type='user', payload_list=True,
             method='POST',
             allowed_param=['user_id', 'screen_name', 'include_entities', 'tweet_mode']
-        )
+        )(post_data=post_data)
 
-    def me(self):
+    def me(self, *args, **kwargs):
```
Harmon758, line 339: I don't think this is necessary.

tweepy/api.py
```diff
@@ -798,9 +763,9 @@ def mutes_ids(self):
             payload_type='ids',
             allowed_param=['cursor'],
             require_auth=True
-        )
-    
-    @property
+        )(*args, **kwargs)
+    mutes_ids.pagination_mode = 'cursor'
+
     def mutes(self):
```
Harmon758, line 769:
```suggestion
    def mutes(self, *args, **kwargs):
```

tweepy/api.py
```diff
@@ -812,10 +777,9 @@ def mutes(self):
             allowed_param=['cursor', 'include_entities', 'skip_status'],
             required_auth=True
         )
```
Harmon758, line 779:
```suggestion
        )(*args, **kwargs)
```

Harmon758 (changes requested):
This should make use of the pagination decorator added with https://github.com/tweepy/tweepy/commit/cb3f22af66ddf9ddca3b57ff0d6d24a51902ff42.

There's also conflicts that need to be resolved now.

Harmon758:
This should now be resolved with https://github.com/tweepy/tweepy/commit/ab02eb532018dd2b0043c990cd2b3c12fb9ca6df.

Issue #1197, opened 2019-04-25
Add API.mutes for mutes/users/list endpoint


Harmon758:
[The mutes/users/list endpoint](https://developer.twitter.com/en/docs/accounts-and-users/mute-block-report-users/api-reference/get-mutes-users-list) was left out from #1055.

Harmon758:
Resolved by https://github.com/tweepy/tweepy/commit/97b64015351201ef42f8c67134cd4180f81f5b7f (#1215) and https://github.com/tweepy/tweepy/commit/29bae1fde3ae0636fa13e5ed69f68f2af75c110b.

PR #1215, opened 2019-05-10
Add API.mutes

blcksrx:
Add API.mutes_list for mutes/users/list endpoint #1197

### DIFF
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -799,6 +799,20 @@ class API(object):
             allowed_param=['cursor'],
             require_auth=True
         )
+    
+    @property
+    def mutes(self):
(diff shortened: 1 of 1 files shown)


tweepy/api.py
```diff
@@ -843,7 +857,7 @@ def blocks(self):
 
     @property
     def blocks_ids(self):
-        """ :reference: https://developer.twitter.com/en/docs/accounts-and-users/mute-block-report-users/api-reference/get-blocks-ids
```
okomarov: @blcksrx I guess this change is unvoluntary.

blcksrx: Ah yes! I'm going to fix it now!

PR #1255, opened 2019-08-07
Drop Python 2 support

### LINKED ISSUE #1253: Drop Python 2 support
With [Python 2.7 reaching its EOL at the end of the year](https://www.python.org/dev/peps/pep-0373/), I'd like to start planning to drop support for it.
It would increase maintainability and reduce a lot of the overhead required for compatibility, as well as allowing improvements that would be difficult to implement while supporting Python 2.

Specifically, I want to look into adding asynchronous support with asyncio (#732), i.e. async/await syntax, and maybe with aiohttp (or requests 3 if that's released soon), especially to replace the current threading method for streams.

I think the current plan should be for Tweepy ~~v3.9~~ (v3.9 was released without noting that it would be the last version to support Python 2 in its changelog and release notes, so a v3.10 is planned) v3.10 to be the last version to support Python 2.7, and then for v4.0 to be Python ~~3.5+~~ (3.5 is also EOL now) 3.6+.

hugovk:
Fixes #1253.


Harmon758:
This PR is a little premature. As I said in #1253, I expect at least another minor version release (noting in the changelog/release notes that it'll be the last version to support Python 2) beforehand.

There was also a reason I self-assigned the issue. I'm not particularly keen on using an automated tool to do this (particularly one that, to be honest, I've never even heard of) in a single commit. This also doesn't resolve the issue entirely, as asyncio support is a key part of the reasoning for dropping Python 2 support.

As of now, I haven't really reviewed the PR, but I'm not sure if it'll be used in the future, as it'll likely be superseded by my own next year.

hugovk:
That's fine, sorry for being early! Feel free to close this now, or later if you wish.

Harmon758:
No problem. Thanks for understanding.

PR #1256, opened 2019-08-07
Test on Python 3.8 beta to avoid surprises

hugovk:
Python 3.8 is in beta and due for release on 2019-10-21.

* https://www.python.org/dev/peps/pep-0569/#schedule

Now is a good to time to start testing on Python 3.8 to make sure there are no surprises when it is released.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ matrix:
   include:
     - python: 3.7
       dist: xenial
+    - python: 3.8-dev
+      dist: xenial
 
 env:
   global:


Harmon758:
I've been testing Tweepy with Python 3.8 since the alpha releases and with each beta release, and there haven't been any issues. I don't think any of the changes in Python 3.8 affect Tweepy, and I don't think it's necessary to add it to CI yet.

When it does release, I expect to be able to promptly add it as a supported version, and if there are any issues at that point, I'll handle them then.

hugovk:
Sounds good, thanks!

PR #1286, opened 2019-10-26
added a virtual environment for mac users

subodhk01:


Harmon758:
This is not something that should be part of the library.
If you're testing out pull requests, please use your own repository to do so.

PR #1086, opened 2018-08-04
Allow image filenames without extension

shuuji3:
Hello, 

This PR fix the issue #1060. 
This allows uploading image files without extension such as `.gif`, `.jpg`, `.jpeg`, or `png`.

PR #1297, opened 2019-11-24
🐛 Bugfix when checking None mimetypes

catleeball:
When mimetypes.guess_type encounters a type it is unfamiliar with,
it returns the tuple (None, None). In api.py, when we check for an
unrecognized filetype, it is before the assignment
`file_type = file_type[0]`. As a result, this block always passes,
even when an unrecognized filetype is encountered.

This change hoists the file_type assignment to the first value of the
tuple a few lines higher to remedy the issue.

### DIFF
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -1379,9 +1379,9 @@ class API(object):
 
         # image must be gif, jpeg, or png
         file_type = mimetypes.guess_type(filename)
+        file_type = file_type[0]
         if file_type is None:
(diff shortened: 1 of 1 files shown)


MichaelCurrin:
This is no longer relevant - on master the flow is now different.

```
        file_type = imghdr.what(filename)
        if file_type is None:
```

catleeball:
Ah, no worries! I'll close this PR.

Harmon758:
Stale due to #1086

PR #1306, opened 2019-12-24
make User object hashable to allow use with sets

mef51:
this allows creating sets out of the results of queries and enables useful set operations and one-liners like 
```connections = list(set(followers) & set(friends))```

### CODE REVIEW

File: tweepy/models.py [hunk 1]
```diff
@@ -224,7 +224,12 @@ def __ne__(self, other):
             return result
 
         return not result
-
+    
+    def __hash__(self):
+        if hasattr(self, 'id'):
+            return hash(self.id)
```
Harmon758, line 230: Using `hash` is unnecessary here, since `User.id` is already an integer.
```suggestion
            return self.id
```

Harmon758 (changes requested):
I think it should be safe to assume `User.id` exists, as in `__eq__`.

tweepy/models.py
```diff
@@ -211,6 +211,12 @@ def __eq__(self, other):
 
         return NotImplemented
 
+    def __hash__(self):
+        if hasattr(self, 'id'):
+            return self.id
+        else:
+            raise TypeError('unhashable type: {} (no id attribute)'.format(type(self)))
```
Harmon758, line 218:
```suggestion
        return self.id
```

Harmon758 (changes requested):

Harmon758 (approved):

mef51:
Thanks!

PR #1319, opened 2020-01-17
Use pip's git+url syntax for installs

PythonCoderAS:
This is better, as it does not use the easy install program that `python setup.py install` uses

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,12 @@ GitHub and install it manually:
     cd tweepy
     python setup.py install
 
+You may also install Tweepy directly from the GitHub 
+repository using pip:
+
+    pip install git+https://github.com/tweepy/tweepy.git
+
+
 Python 2.7, 3.5, 3.6, 3.7, & 3.8 are supported.
 
 Community
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -7,7 +7,4 @@ Install from PyPI::
 
 Install from source::
 
-    git clone git://github.com/tweepy/tweepy.git
-    cd tweepy
-    python setup.py install
-
+    pip install git+https://github.com/tweepy/tweepy.git


README.md
```diff
@@ -17,9 +17,7 @@ is by using pip to pull it from PyPI:
 You may also use Git to clone the repository from
 GitHub and install it manually:
 
-    git clone https://github.com/tweepy/tweepy.git
-    cd tweepy
-    python setup.py install
+    pip install git+https://github.com/tweepy/tweepy.git
```
Harmon758: That doesn't match the description anymore.
```suggestion
    git clone https://github.com/tweepy/tweepy.git
    cd tweepy
    pip install .
```

PythonCoderAS: This would literally negate the only change.

Harmon758: The relevant part of the change is using pip to install the library rather than easy_install, which this still does.

PythonCoderAS: I thought that doing `pip install git+...` would use pip for it

Harmon758: Yes, obviously, but again, that doesn't match the description for what this is supposed to do.

PythonCoderAS: I changed the content of the message.

README.md
```diff
@@ -14,12 +14,9 @@ is by using pip to pull it from PyPI:
 
     pip install tweepy
 
-You may also use Git to clone the repository from
-GitHub and install it manually:
+You may also install tweepy from the Git repo:
```
Harmon758:
```suggestion
You may also install Tweepy from the GitHub repository:
```

docs/install.rst
```diff
@@ -7,7 +7,4 @@ Install from PyPI::
 
 Install from source::
 
-    git clone git://github.com/tweepy/tweepy.git
-    cd tweepy
-    python setup.py install
-
+    pip install git+git://github.com/tweepy/tweepy.git
```
Harmon758: The scheme should be consistent with the README.

README.md
Harmon758: There's no reason to replace the instructions for cloning the repository.
You can simply add additional instructions for installing from GitHub without cloning the repository.

Issue #1336, opened 2020-02-28
media_upload restricts gif size


achasveachas:
It looks like the `media_upload` function restricts the size for all media uploads to 4883kb ([here](https://github.com/tweepy/tweepy/blob/3733fd673b04b9aa193886d6b8eb9fdaf1718341/tweepy/api.py#L222-L223))

Though, according to the [Twitter API docs](https://developer.twitter.com/en/docs/media/upload-media/overview) the limit for gifs and videos is 15MB.

Can the 5MB limit be conditioned on the media type?

PR #1338, opened 2020-03-01
Increase max image size for gif

achasveachas:
Raises the maximum size of gifs allowed to 15MB per [Twitter's documentaion](https://developer.twitter.com/en/docs/media/upload-media/overview).

This resolves https://github.com/tweepy/tweepy/issues/1336

**Note:** I set the max size to 15360KB (15 MB). I noticed that the max size for images was slightly less than 5 MB though (4883KB). I couldn't find the reason for that, but if there is one I can revise the size for gifs as well.

### DIFF
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -222,7 +222,7 @@ class API(object):
 
         file_type = imghdr.what(filename)
         if file_type == 'gif':
-            max_size = 14648
+            max_size = 14649
         else:
             max_size = 4883
 
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -219,7 +219,14 @@ class API(object):
             :allowed_param:
         """
         f = kwargs.pop('file', None)
-        headers, post_data = API._pack_image(filename, 4883,
+
+        file_type = imghdr.what(filename)
+        if file_type == 'gif':
+            max_size = 15360
+        else:
+            max_size = 4883
+
+        headers, post_data = API._pack_image(filename, max_size,
                                              form_field='media', f=f)
         kwargs.update({'headers': headers, 'post_data': post_data})
 
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -227,7 +227,8 @@ class API(object):
             max_size = 4883
 
         headers, post_data = API._pack_image(filename, max_size,
-                                             form_field='media', f=f)
+                                             form_field='media', f=f,
+                                             file_type=file_type)
         kwargs.update({'headers': headers, 'post_data': post_data})
 
         return bind_api(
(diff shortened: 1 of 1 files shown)


MichaelCurrin:
I noticed this PR because a similar issue #1359 was just brought up.

---

@achasveachas The changes look good, except for the number. If you can adjust that then I'd approve. See notes below.

The function you modified does a call to `_pack_image` which does a calc given a value - `max_size * 1024` [https://github.com/tweepy/tweepy/blob/3733fd6/tweepy/api.py#L1364](api.py#L1364). And so using `max_size` as `4883` gets to `5,000,192` - which is slightly _over_ 5MB and not under as you mentioned.

Reverse engineering that:

```
5MB
= 5*000*1000/1024
= 4882.8125
~= 4883
```

Therefore:
```
15MB
= 15*1000*1000/1024
= 14648.4375
~= 14648  # rounded up to be safe
```

Therefore `14648` is your magic number ⭐  🚀 

---

I see you have a doc reference already, but here's another extra reference I found: https://developer.twitter.com/en/docs/media/upload-media/uploading-media/media-best-practices

> Image size <= 5 MB, animated GIF size <= 15 MB

tweepy/api.py
```diff
@@ -219,7 +219,14 @@ def media_upload(self, filename, *args, **kwargs):
             :allowed_param:
         """
         f = kwargs.pop('file', None)
-        headers, post_data = API._pack_image(filename, 4883,
+
+        file_type = imghdr.what(filename)
+        if file_type == 'gif':
+            max_size = 15360
```
MichaelCurrin, line 225:
```suggestion
            max_size = 14648
```

achasveachas:
@MichaelCurrin that fix is in.

tweepy/api.py
achasveachas: Done

tweepy/api.py
```diff
@@ -219,7 +219,14 @@ def media_upload(self, filename, *args, **kwargs):
             :allowed_param:
         """
         f = kwargs.pop('file', None)
-        headers, post_data = API._pack_image(filename, 4883,
+
+        file_type = imghdr.what(filename)
+        if file_type == 'gif':
+            max_size = 14648
```
Harmon758, line 225: This should actually be 14649. @MichaelCurrin actually rounded 14648.4375 down to 14648 rather than rounding up to 14649 as it says. This would cause GIFs between 14,999,553 (14648 * 1024 + 1) and 15,000,000 bytes (inclusive) to error even though Twitter's API should accept them.
```suggestion
            max_size = 14649
```

achasveachas: Whoops! Always check the math :wink: 

Fixed.

achasveachas:
I added `file_type` as an optional argument to `API._pack_image`. Now it will only determine the file_type in cases where it isn't passed in.

Harmon758:
Thanks!

PR #1342, opened 2020-03-23
EHN: Save the raw JSON on DirectMessages as well.

musically-ut:
This brings the behavior of the `DirectMessage` model in line with other `Model`s.

### DIFF
--- a/tweepy/models.py
+++ b/tweepy/models.py
@@ -231,9 +231,9 @@ class DirectMessage(Model):
     @classmethod
     def parse(cls, api, json):
         dm = cls(api)
-        setattr(dm, '_json', json)
         if "event" in json:
             json = json["event"]
+        setattr(dm, '_json', json)
         for k, v in json.items():
             setattr(dm, k, v)
         return dm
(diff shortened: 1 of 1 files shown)


tweepy/models.py
```diff
@@ -231,6 +231,7 @@ class DirectMessage(Model):
     @classmethod
     def parse(cls, api, json):
         dm = cls(api)
+        setattr(dm, '_json', json)
         if "event" in json:
             json = json["event"]
```
Harmon758, line 236: This should probably be after `json` is set to the specific event that [GET direct_messages/events/show](https://developer.twitter.com/en/docs/direct-messages/sending-and-receiving/api-reference/get-event) (`API.get_direct_messsage`) returns, so that the JSON represents the actual event and so as to be consistent with the JSON that would be representing a Direct Message event when returned by [GET direct_messages/events/list](https://developer.twitter.com/en/docs/direct-messages/sending-and-receiving/api-reference/list-events) (`API.list_direct_messages`):
```suggestion
        if "event" in json:
            json = json["event"]
        setattr(dm, '_json', json)
```

Harmon758:
Thanks!

PR #1347, opened 2020-04-02
Update streaming_how_to.rst

### LINKED ISSUE #1346: Tweepy documentation does not specify that OAuth 1a is required for Streaming
I have been reading the [Tweepy documentation](http://docs.tweepy.org/en/latest/index.html) as I try to use it for a new project.

As it wasn't clear to me what the difference between OAuth 1a and OAuth 2 authentication mechanisms is, I opted to use OAuth 2 as it involved fewer steps. Great success followed, I was able to search for users, hashtags, etc.

However, when trying to stream, I was consistently getting 401 errors from Twitter. There are existing issues surrounding Tweepy and 401 errors, such as #877.

Eventually I decided to try another python library for streaming, as I had already double checked that the system time was set to UTC and synced with NTP with a drift of <5 minutes. That's when I started reading [Twython's documentation](https://twython.readthedocs.io/en/latest/usage/streaming_api.html) and stumbled across this:

> The Streaming API requires that you have OAuth 1 authentication credentials.

I changed my Tweepy authentication to use OAuth 1a, and magically the Streaming API works.

I went back and reviewed Tweepy's [authentication documentation](http://docs.tweepy.org/en/latest/auth_tutorial.html) and [streaming documentation](http://docs.tweepy.org/en/latest/streaming_how_to.html), and nowhere do I see it mentioned that Streaming is only possible using OAuth 1a authentication.

It would be very nice to add this warning to your documentation, as Twython has done. Knowing this in advance would have saved me untold hours of debugging and hair pulling.

halmartin:
PR to resolve #1346

### DIFF
--- a/docs/streaming_how_to.rst
+++ b/docs/streaming_how_to.rst
@@ -15,6 +15,9 @@ by offering a first walk through.  Some features of Tweepy streaming are
 not covered here. See streaming.py in the Tweepy source code. 
 
 API authorization is required to access Twitter streams. 
+
+**Important**: You must authenticate using OAuth 1a if you want to use the streaming API.
+
 Follow the :ref:`auth_tutorial` if you need help with authentication. 
 
 Summary
--- a/docs/streaming_how_to.rst
+++ b/docs/streaming_how_to.rst
@@ -16,7 +16,7 @@ not covered here. See streaming.py in the Tweepy source code.
 
 API authorization is required to access Twitter streams. 
 
-**Important**: You must authenticate using OAuth 1a if you want to use the streaming API.
+**Important**: You must authenticate using _OAuth 1a_ to get a token with _user_ context if you want to use the streaming API. This can be done using an App Access Token as on Getting Started page, or using the longer user-login flow.
 
 Follow the :ref:`auth_tutorial` if you need help with authentication. 
 
@@ -125,4 +125,3 @@ Documentation`_. ::
           # returning non-False reconnects the stream, with backoff.
 
 For more information on error codes from the Twitter API see `Twitter Response Codes Documentation`_.
-


MichaelCurrin:
@halmartin 


I think the problem is that one needs to do auth with a user context (whether using access credentials OR using a user-login flow). Since the application-only flow (as on the [OAuth2 page](http://docs.tweepy.org/en/latest/auth_tutorial.html#oauth-2-authentication) of Twitter docs) works great for searching but **not** for streaming, according your issue.

I don't think your comment going to make sense to the average user for what to do next.

Can I suggest a link to the Getting Started page added here, as that auth case satisfies this need. http://docs.tweepy.org/en/latest/getting_started.html

Or elaborate and say that one has to use CONSUMER credentials _and_ ACCESS credentials.

MichaelCurrin:
I saw the Tweepy docs on _OAuth 1a_ are long and only cover the user-login process http://docs.tweepy.org/en/latest/getting_started.html

so that's **not** so useful to link. As it means one has to user the browser to sign in as a user, while using the flow in Getting Started is the more common usecase and appropriate for streaming (that's what I've used for streaming before and I think that is the approach you are using too).

halmartin:
@MichaelCurrin My main issue is that the Tweepy [streaming documentation](https://docs.tweepy.org/en/latest/streaming_how_to.html) states the following:
> API authorization is required to access Twitter streams. Follow the Authentication Tutorial if you need help with authentication.

**Nowhere** does it state in the [Authentication Tutorial](https://docs.tweepy.org/en/latest/auth_tutorial.html) that you must use OAuth 1 to stream.

> OAuth 2 is a method of authentication where an application makes API requests without the user context. Use this method if you just need read-only access to public information.

This doesn't make it obvious to the reader that OAuth 2 is unsuitable for streaming. In my interpretation, the Twitter API stream is "read-only access to public information"

The warning about using OAuth 1 for streaming is plainly obvious in [twython's streaming documentation](https://twython.readthedocs.io/en/latest/usage/streaming_api.html):
> The Streaming API requires that you have OAuth 1 authentication credentials. If you don’t have credentials, head over to the authentication section and find out how!

I am completely indifferent as to where the notice of using OAuth 1 is placed in the Tweepy documentation. But, IMO, the most logical place to put it is the Streaming instructions (or both the Streaming instructions and the Authentication tutorial). If someone is reading the documentation sequentially, the Authentication Tutorial comes before streaming, so to avoid wasted effort it might be helpful to add a notice there as well.

As you also point out, OAuth 1 is significantly more effort to use, involving multiple steps and user interaction. It was for that very reason that I initially decided to just use OAuth 2 for streaming (why do more work if you don't have to?). That decision cost me days of debugging, which was only resolved when I gave up on Tweepy and started reading the Twython documentation during my reimplementation.

docs/streaming_how_to.rst
```diff
@@ -15,6 +15,9 @@ by offering a first walk through.  Some features of Tweepy streaming are
 not covered here. See streaming.py in the Tweepy source code. 
 
 API authorization is required to access Twitter streams. 
+
+**Important**: You must authenticate using OAuth 1a if you want to use the streaming API.
```
MichaelCurrin, line 19:
```suggestion
**Important**: You must authenticate using _OAuth 1a_ to get a token with _user_ context if you want to use the streaming API. This can be done using an App Access Token as on Getting Started page, or using the longer user-login flow.
```

MichaelCurrin:
I agree with putting your comment in where it is for streaming tutorial.

> Follow the :ref:`auth_tutorial` if you need help with authentication.

Which means someone will click through there and get misled into thinking they **must** use the user login flow. But that is not the case. 

The `.OAuthHandler` on the **3-line** authentication on Getting Started page uses OAuth 1a and it find it works great for streaming.

MichaelCurrin:
Unfortunately, the auth page in Tweepy docs are misleading as it only covers one type (user-login) and not the App Access Token flow (as on Getting Started) which also has user context.

To be clear on naming, the one _without_ user context is Application-Only Token, which is the on under OAuth2 in docs

rafaelnovello:
> PR para resolver [# 1346](https://github.com/tweepy/tweepy/issues/1346)

Thank you very much! Your issue and PR help me to finally start consuming twitter streaming data! =D

seeess1:
I'm currently getting 401 errors when I try to run my tweepy streaming script on a GCP VM but it works just fine when I run it locally (with the same auth credentials). Based on what I'm describing, does it sound like this could be an OAuth 1 vs OAuth 2 issue? Or is it probably something else?

Harmon758:
This is stale now, with https://github.com/tweepy/tweepy/commit/bf641965278d2486cebef45bb940e2d2e691d420.

@seeess1 That's unrelated to this PR.
The description of a 401 error code from the API documentation:
> HTTP authentication failed due to either:
> - Invalid basic auth credentials, or an invalid OAuth request.
> - Out-of-sync timestamp in your OAuth request (the response body will indicate this).
> - Too many incorrect passwords entered or other login rate limiting.

For more information, see https://developer.twitter.com/en/docs/tweets/filter-realtime/guides/connecting and https://developer.twitter.com/en/docs/basics/authentication/guides/authorizing-a-request.

You should check that your system clock is in sync with NTP.

PR #1349, opened 2020-04-05
Fixed broken encoding for track param for Stream

creotiv:
By default, requests use the quote_plus() encoding method which breaks Twitter API filtering with track param.
So instead + we need to use %20 encoding for space.

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -267,7 +267,7 @@ class Stream(object):
                 resp = self.session.request('POST',
                                             url,
                                             data=urllib.parse.urlencode(
-                                                self.body, quote_via=urllib.parse.quote),,
+                                                self.body, quote_via=urllib.parse.quote),
                                             timeout=self.timeout,
                                             stream=True,
                                             auth=auth,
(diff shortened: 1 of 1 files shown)


MichaelCurrin:
The code you wrote only work for PY3.

The namespace is different for PY2, so `six.moves.urllib` is safer to use.

If you _do_ want to drop PY2 support with your code as is, be expicit about this like I mentioned above here. The maintainers may not want to drop support just yet and so `six` is needed, or maybe your approach is fine.

Also, do you think your code will work as intended if you collapse your nested function call into a single call to `urllib.parse.quote_plus` ?

Some unit test would be good too to capture the intent of converting even if you have to move your logic to a new function so it can be tested alone. Though quote_plus itself will probably have good coverage in its original library

Harmon758:
> which breaks Twitter API filtering with track param

This shouldn't be the case, and I'm unable to reproduce any issues with spaces in phrases for the `track` parameter of `Stream.filter`. Feel free to create a new issue if you have a reproducible example of this breakage.

PR #1351, opened 2020-04-11
Passing error code to RateLimitError for backwards compatibility with TweepError

jefigo:
As it's right now, whenever an API operation hits the rate limit Tweepy raises a `tweepy.error.RateLimitError`, although this error is meant to be fully backwards compatible with `TweepError` the `api_code` property is always null. 

With this pull request merged, a `tweepy.error.RateLimitError` will include the error code in the `api_code` property. 

This pull request also makes easier to check if the error was raised because the operation hit the rate limit by passing to `tweepy.error.is_rate_limit_error_message` the api error code previously parsed.

### DIFF
--- a/tweepy/binder.py
+++ b/tweepy/binder.py
@@ -228,8 +228,8 @@ def bind_api(**config):
                     error_msg = "Twitter error response: status code = %s" % resp.status_code
                     api_error_code = None
 
-                if is_rate_limit_error_message(error_msg):
-                    raise RateLimitError(error_msg, resp)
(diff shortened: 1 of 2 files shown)


MichaelCurrin:
Using the result of `parser.parse_error` to get the error code here looks good as it means the code is dry and shorter. Haven't run this but it looks sound to me.

Harmon758:
This is stale now, with https://github.com/tweepy/tweepy/commit/cd5f696d09530f86ac0edf1ec0fe0a02578a3920.

PR #1364, opened 2020-05-15
add quick_reply_options

om8007:
### add quick_reply_options for passing options object for _quick_reply_type = 'options'_
adding **'options'** in **message_data['quick_reply']** allows sender to pass options object for **quick_reply_type = 'options'**. Without 'options' it raises _tweepy.error.TweepError: [{'code': 214, 'message': 'event.message_create.message_data.quick_reply: quick_reply per-type node not found for quick_reply type options'}]_

### DIFF
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -427,7 +427,7 @@ class API(object):
             require_auth=True
         )
 
-    def send_direct_message(self, recipient_id, text, quick_reply_type=None,
+    def send_direct_message(self, recipient_id, text, quick_reply_type=None, quick_reply_options = None,
                             attachment_type=None, attachment_media_id=None):
         """
         Send a direct message to the specified user from the authenticating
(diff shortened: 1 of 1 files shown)


Harmon758 (changes requested):
Location and text input quick replies should be supported as well.

Harmon758:
Sorry, due to [Location and Text Input Quick Replies still being documented](https://twittercommunity.com/t/retired-location-and-text-input-quick-replies-are-still-documented/148102), I didn't realize that they had been retired from the API.

However, since Options Quick Replies are the only type now, `quick_reply_options` can simply replace `quick_reply_type` as a parameter, with the type set to "options" if options are passed, and with this PR being stale, I've gone ahead and done that with https://github.com/tweepy/tweepy/commit/d8b02de5e27c91b477cc699a064507201f53a139.

PR #1371, opened 2020-05-21
Remove null characters from the response payload

jefigo:


MichaelCurrin:
@jefigo is a there an issue you can add or some output or screen to show the problem and the result to ensure this works as expected?

I was wondering if null character needs only one backslash otherwise you are escaping the backlash

Also there are other ways to make a null character from quick research.

```
"\xc0"
"\0"
```

Also is the null possibly only at the _end_ which could be resolved with this? `text.strip()`

Harmon758:
I'm not sure why this is necessary. What's the motivation behind this PR?

PR #1438, opened 2020-09-20
Added rate limit advice to cursor tutorial.

SolusTheAussie:
Added a section to the cursor tutorial advising the reader that the rate-limits can be optimised by setting the count value.


Harmon758:
Thanks, but this is now stale with / superseded by https://github.com/tweepy/tweepy/commit/652fece7f78f90ea11f206b4a045a2450db42cc9.

PR #1447, opened 2020-10-02
Introducing async (asynchronous) to on_data function

refekt:
Introducing `async` to allow the ability to be non-blocking.

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -41,7 +41,7 @@ class StreamListener(object):
         """
         pass
 
-    def on_data(self, raw_data):
+    async def on_data(self, raw_data):
         """Called when raw data is received from connection.
 
         Override this method if you wish to manually handle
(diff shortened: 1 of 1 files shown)


joshthecoder:
Looks good. Merge on this is blocked by us removing 2.7 from our support matrix (which is happening soon given it is EOL).

joshthecoder (approved):

Harmon758:
This is a breaking change that adds little to no value. The entirety of the processing and handling of the stream is still blocking here, including the request and connection itself, not to mention `Stream.sample` and the other `StreamListener` methods, which aren't included with this. In fact, this breaks `Stream.sample`, since it wouldn't be awaiting `Stream._start`. Not only that, but it breaks the actual existing asynchronous functionality in place with the `is_async` parameter that uses threading, since the thread would be improperly passed a coroutine. Simply changing a few methods to coroutines does not introduce asynchronous support or support for asyncio. For that, a different library has to be used for the request and connection itself, since the requests library is blocking, and most of the logic has to be rewritten to use that library and actually be non-blocking. See #1491.

Issue #1466, opened 2020-12-09
api.update_profile_background_image returns "Sorry, that page does not exist"


TallGorilla:
Im creating a script that will change the user banner, the point is that **every time it try to change the image** by using `api.update_profile_background_image("the/image/location.png")` it returns `Exception has occurred: TweepError
[{'message': 'Sorry, that page does not exist', 'code': 34}]`

_**_How can i fix this?_**_

Thanks in advance.

TallGorilla:
Closed as i recenlty discovered there is the update_profile_banner option XD

Harmon758:
The [POST account/update_profile_background_image endpoint](https://web.archive.org/web/20190809210233/https://developer.twitter.com/en/docs/accounts-and-users/manage-account-settings/api-reference/post-account-update_profile_background_image) that `API.update_profile_background_image` used has been deprecated and removed, since background profile images have not been available or visible on Twitter.com since July 2015.
https://twittercommunity.com/t/upcoming-changes-to-the-developer-platform/104603

I've gone ahead and removed the method with https://github.com/tweepy/tweepy/commit/8dba191518366fb756440302eff86d5a868e0306.

PR #1467, opened 2020-12-11
Updating banner frontend

TallGorilla:
The banner front end has been changed to https://api.twitter.com/1.1/account/update_profile_banner.json as you can see here: https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/manage-account-settings/api-reference/post-account-update_profile_banner

### DIFF
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -708,7 +708,7 @@ class API(object):
         headers, post_data = API._pack_image(filename, 800, f=f)
         return bind_api(
             api=self,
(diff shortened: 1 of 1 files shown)


TallGorilla:
I just noticed that this creates a new bug beacouse to interact with that new frontend you have to use the argument "banner" instead of filename.

Working on it.

Harmon758:
`API.update_profile_banner` already exists, and `API.update_profile_background_image` has been removed now.
See #1466.

PR #1413, opened 2020-08-10
Fixes #1412

fkropfhamer:
Fixes #1412 by transforming file argument to bytes and passing to `imghdr.what` in `api.media_upload`

PR #1475, opened 2020-12-19
Allows media upload from BytesIO

keatonconrad:
Allows the `media_upload()` and `_pack_image()` functions to guess the file type from BytesIO data. All this does is pass in the already-existing `file`/`f` parameters to `imghdr` to guess the file type and allow the data to be processed as if it were a file.

### DIFF
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -225,6 +225,8 @@ class API(object):
             file_type = imghdr.what(filename) or mimetypes.guess_type(filename)[0]
         else:
             file_type = imghdr.what(filename, h=f.read()) or mimetypes.guess_type(filename)[0]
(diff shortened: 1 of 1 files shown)


Harmon758 (changes requested):
Since this reads the entire file to pass it to `imghdr.what` and the file object's position isn't reset, an empty string ends up being uploaded as the image.

Harmon758 (changes requested):
#1413 has been merged now, so this PR now has conflicts that should be resolved in favor of the changes to `media_upload` from that PR.

Harmon758:
Since this is one of the last remaining PRs I want to include for v3.10, I've gone ahead and resolved the conflicts.

Harmon758 (approved):

PR #1484, opened 2020-12-26
Doc fix

dplefevre:
Updated documentation on user_timeline API method: removed 'page' parameter, reordered allowed params to match Twitter API documentation, and added description for 'include_rts'


docs/api.rst
```diff
@@ -85,7 +80,7 @@ Timeline methods
 
 
 .. method:: API.user_timeline([id/user_id/screen_name], [since_id], [max_id], \
-                              [count], [page])
+                              [count])
```
Harmon758, line 83:
```suggestion
.. method:: API.user_timeline([id/user_id/screen_name], [since_id], [count], \
                              [max_id], [trim_user], [exclude_replies], \
                              [include_rts])
```

docs/api.rst
```diff
@@ -95,9 +90,16 @@ Timeline methods
    :param user_id: |user_id|
    :param screen_name: |screen_name|
    :param since_id: |since_id|
-   :param max_id: |max_id|
    :param count: |count|
-   :param page: |page|
+   :param max_id: |max_id|
+   :param trim_user: |trim_user|
+   :param exclude_replies: |exclude_replies|
+   :include_rts: When set to false , the timeline will strip any native
```
Harmon758, line 97:
```suggestion
   :param include_rts: When set to ``false``, the timeline will strip any native
```

docs/parameters.rst
```diff
@@ -4,6 +4,7 @@
 .. |cursor| replace:: Breaks the results into pages. Provide a value of  -1 to begin paging. Provide values as returned to in the response body's next_cursor and previous_cursor attributes to page back and forth in the list.
 .. |date| replace:: Permits specifying a start date for the report. The date should be formatted YYYY-MM-DD.
 .. |exclude| replace:: Setting this equal to hashtags will remove all hashtags from the trends list.
+.. |exclude_replies|:: This parameter will prevent replies from appearing in the returned timeline. Using ``exclude_replies`` with the ``count`` parameter will mean you will receive up-to count Tweets — this is because the ``count`` parameter retrieves that many Tweets before filtering out retweets and replies.
```
Harmon758, line 7:
```suggestion
.. |exclude_replies| replace:: This parameter will prevent replies from appearing in the returned timeline. Using ``exclude_replies`` with the ``count`` parameter will mean you will receive up-to count Tweets — this is because the ``count`` parameter retrieves that many Tweets before filtering out retweets and replies.
```

Harmon758 (changes requested):
`id` should also be removed as an allowed parameter.

dplefevre:
It seems that after removing the ID parameter, test_api.py/testusertimeline() is failing. I have not used nosetest before so I am having a hard time understanding the exact issue at a glance. I am sure I can figure it out if I take some time and look into it, though. Would you like me to try to update the test as well?

Harmon758:
No problem, I can update the tests.

Harmon758 (approved):

Issue #1489, opened 2020-12-30
API.verify_credentials() throws Error instead of returning False if consumer_key is set to None


VenNeptury:
Title says it all. 

**Example code:**
````py
import tweepy

auth = tweepy.OAuthHandler(None, None)

api = tweepy.API(auth)

api.verify_credentials()
````

<details>
  <summary> Traceback</summary>
  
````
Traceback (most recent call last):
  File "/home/ven/.local/lib/python3.8/site-packages/tweepy/binder.py", line 184, in execute
    resp = self.session.request(self.method,
  File "/usr/lib/python3/dist-packages/requests/sessions.py", line 519, in request
    prep = self.prepare_request(req)
  File "/usr/lib/python3/dist-packages/requests/sessions.py", line 452, in prepare_request
    p.prepare(
  File "/usr/lib/python3/dist-packages/requests/models.py", line 317, in prepare
    self.prepare_auth(auth, url)
  File "/usr/lib/python3/dist-packages/requests/models.py", line 548, in prepare_auth
    r = auth(self)
  File "/home/ven/.local/lib/python3.8/site-packages/requests_oauthlib/oauth1_auth.py", line 108, in __call__
    r.url, headers, _ = self.client.sign(
  File "/usr/lib/python3/dist-packages/oauthlib/oauth1/rfc5849/__init__.py", line 313, in sign
    ('oauth_signature', self.get_oauth_signature(request)))
  File "/usr/lib/python3/dist-packages/oauthlib/oauth1/rfc5849/__init__.py", line 127, in get_oauth_signature
    uri, headers, body = self._render(request)
  File "/usr/lib/python3/dist-packages/oauthlib/oauth1/rfc5849/__init__.py", line 209, in _render
    headers = parameters.prepare_headers(
  File "/usr/lib/python3/dist-packages/oauthlib/oauth1/rfc5849/utils.py", line 32, in wrapper
    return target(params, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/oauthlib/oauth1/rfc5849/parameters.py", line 59, in prepare_headers
    escaped_value = utils.escape(value)
  File "/usr/lib/python3/dist-packages/oauthlib/oauth1/rfc5849/utils.py", line 56, in escape
    raise ValueError('Only unicode objects are escapable. ' +
ValueError: Only unicode objects are escapable. Got None of type <class 'NoneType'>.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #1490, opened 2020-12-30
Make API.verify_credentials() return False instead of throwing if consumer_key is None

VenNeptury:
Fixes tweepy/tweepy#1489

### DIFF
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -664,6 +664,10 @@ class API:
         """ :reference: https://developer.twitter.com/en/docs/accounts-and-users/manage-account-settings/api-reference/get-account-verify_credentials
             :allowed_param: 'include_entities', 'skip_status', 'include_email'
         """
+
+        if self.auth.consumer_key is None:
+            return False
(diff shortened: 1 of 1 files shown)


Harmon758:
#1489 is an issue with the error that occurs when signing any API request, not just with `API.verify_credentials`.

VenNeptury:
Regardless of that, why should this method throw instead of returning False as expected? Its purpose is to verify your credentials and since no consumer key is provided, the credentials are invalid. Thus the return should be False in my opinion and not an Exception. On all other methods across the library it makes sense to throw but not on this one.

Harmon758:
Because this method is no different from any other `API` method. It simply makes a request to one of Twitter's API endpoints.
It doesn't need to be special cased.

I've resolved #1489 so that a `TypeError` is raised when initializing `OAuthHandler` with `consumer_key` and/or `consumer_secret` arguments that aren't `str` or `bytes`, so it should never reach the point of calling any `API` method.

PR #1584, opened 2021-05-12
Remove async from filter and sample functions

crunis:
I think `filter` and `sample` functions shouldn't be `async` since they only return a task.

This way, we can directly execute the task:
```
await AsyncStream.filter()
```

If the `async` is left, it forces us to use an initial `await` to retrieve the task and then another `await` to execute it:
```
task = await AsyncStream.filter()
await task
```

### DIFF
--- a/tweepy/asynchronous/streaming.py
+++ b/tweepy/asynchronous/streaming.py
@@ -146,9 +146,7 @@ class AsyncStream:
 
     def filter(self, *, follow=None, track=None, locations=None,
                      filter_level=None, languages=None, stall_warnings=False):
-        """|coroutine|
-
-        Filter realtime Tweets
+        """Filter realtime Tweets
 
         Parameters
         ----------
@@ -235,9 +233,7 @@ class AsyncStream:
         return self.task
 
     def sample(self, *, languages=None, stall_warnings=False):
-        """|coroutine|
-
-        Sample realtime Tweets
+        """Sample realtime Tweets
 
         Parameters
         ----------
--- a/tweepy/asynchronous/streaming.py
+++ b/tweepy/asynchronous/streaming.py
@@ -144,7 +144,7 @@ class AsyncStream:
             await self.session.close()
             await self.on_disconnect()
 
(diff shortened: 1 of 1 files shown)


tweepy/asynchronous/streaming.py
```diff
@@ -144,7 +144,7 @@ async def _connect(self, method, endpoint, params={}, headers=None,
             await self.session.close()
             await self.on_disconnect()
 
-    async def filter(self, *, follow=None, track=None, locations=None,
+    def filter(self, *, follow=None, track=None, locations=None,
                      filter_level=None, languages=None, stall_warnings=False):
         """|coroutine|
```
Harmon758, line 149: This documentation indicating it's a coroutine needs to be removed.

tweepy/asynchronous/streaming.py
```diff
@@ -234,7 +234,7 @@ async def filter(self, *, follow=None, track=None, locations=None,
         )
         return self.task
 
-    async def sample(self, *, languages=None, stall_warnings=False):
+    def sample(self, *, languages=None, stall_warnings=False):
         """|coroutine|
 
         Sample realtime Tweets
```
Harmon758, line 240:
```suggestion
        """Sample realtime Tweets
```

Harmon758 (changes requested):
Awaiting the [`asyncio.Task`](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task) is not necessary for it to run. It simply blocks until the Task is done. `await AsyncStream.filter()` already "directly executes" the task (assuming `AsyncStream` is actually an instance of `AsyncStream` here), as does `task = await AsyncStream.filter()`. `await task` simply blocks until the stream disconnects.

However, it's true that `AsyncStream.filter` and `AsyncStream.sample` don't need to be coroutines now, with https://github.com/tweepy/tweepy/commit/59872045009ffb4bb5f42af3938c6a5722fc4f4c, and changing them to not be coroutines would simplify their usage.

crunis:
Updated the documentation of filter and sample functions.

Harmon758 (approved):

Issue #1596, opened 2021-05-30
Documentation site is ahead of current PyPI version


hkennyv:
I'm not sure if this is intended, but the link to the tweepy docs from <https://www.tweepy.org/> link to docs that are up-to-date with the latest commit on the `master` branch.

This is slightly confusing for people who are installing from PyPI: <https://pypi.org/project/tweepy/> since it looks like the package is in the middle of a major version bump so the docs have some API breaking differences (`tweepy.Stream` was the one I ran into).

Ultimately, I was able to find the docs for v3.10.0 here: <https://docs.tweepy.org/en/v3.10.0/> and that's been working for me. Just thought I'd give a heads up to any other people who may run across this.

Harmon758:
I've changed the default Read the Docs version to `stable`.

hkennyv:
@Harmon758 thanks! awesome job on the docs btw, definitely a huge improvement IMO.

PR #1597, opened 2021-05-30
Update documentation for API.rate_limit_status to be consistent with other docstrings

hkennyv:
After running into #1596, I ended up cloning the project and doing a little digging and saw the recent commit history updating the docs.

I found one method that seemed to not be updated (i just did a project search for `:param`). I just updated it to be consistent with the rest of the method docstrings.

Open to any feedback in case I missed anything


Harmon758 (approved):
Ah, I must have missed it at the very end 😅
Thanks!

PR #1621, opened 2021-07-14
fix asyncio timeout error when getting rate limit, status code 420

tuangeek:
improves handle connection errors when asynchronous streaming.

asyncio runs into timeout error if stream reaches rate limit (status_code 420). This is similar to https://github.com/tweepy/tweepy/commit/68e19cc6b9b23d72369ca1520093770eb18a5a9f

### DIFF
--- a/tweepy/asynchronous/streaming.py
+++ b/tweepy/asynchronous/streaming.py
@@ -3,6 +3,7 @@
 # See LICENSE for details.
 
 import asyncio
+from asyncio import TimeoutError
 import json
 import logging
 from math import inf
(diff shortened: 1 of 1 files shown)


tweepy/asynchronous/streaming.py
```diff
@@ -3,6 +3,7 @@
 # See LICENSE for details.
 
 import asyncio
+from asyncio import TimeoutError
```
jai-dewani, line 6: This import won't be required as you have referenced `TimeoutError` as `asyncio.TimeoutError`.

jai-dewani (commented):
I think unnecessary imports can be removed as they are not needed

PR #1623, opened 2021-07-17
Raise NotFound on v2 404 requests

le717:
Raise a `NotFound` exception on requests with a status code of `404` when using `Client`, for consistency with `API` behavior and developer expectations.

### DIFF
--- a/tweepy/client.py
+++ b/tweepy/client.py
@@ -13,8 +13,8 @@ import requests
 import tweepy
 from tweepy.auth import OAuthHandler
 from tweepy.errors import (
-    BadRequest, Forbidden, HTTPException, TooManyRequests, TwitterServerError,
-    Unauthorized
+    BadRequest, Forbidden, HTTPException, NotFound, TooManyRequests,
+    TwitterServerError, Unauthorized
 )
 from tweepy.media import Media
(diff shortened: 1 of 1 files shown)


le717:
@Harmon758 Is there any way you can take a look at this maybe soon? I'm wanting to start looking into using v4a in my apps and it'd be nice to have this in place for me automatically. 😀

Harmon758:
This is most likely fine, but do you have an example of a 404 response?
I think I had the comment there because I wasn't sure if it was a possible response with any of the v2 endpoints.

le717:
It's definitely possible for a 404 response to be raised, when attempting to hit an incorrect endpoint (e.g., `/by/usernam/`). But getting a 404 from a valid but not found exists (e.g, `@does.not.exist`), the result just gets put into the `errors` property on the response. Previously, that would raise `TweepError`, which is the behavior I was expecting here. I made a poor assumption on not much digging, My apologies for the bother. I'll gladly close this is if you think it's not needed.

Harmon758:
I've implemented this now, with https://github.com/tweepy/tweepy/commit/b6b82196d3f0821c184901de985e2cedb56a9db2, to futureproof and support subclassing.

PR #1627, opened 2021-07-23
add except StopIteration + return to limit_handled

AnneliesBotjes:
so it will actually work.

### DIFF
--- a/docs/code_snippet.rst
+++ b/docs/code_snippet.rst
@@ -58,7 +58,10 @@ will wait for 15 minutes each time it hits the rate limit.
                yield next(cursor)
            except tweepy.RateLimitError:
                time.sleep(15 * 60)
-   
+               return
+           except StopIteration:
+               break
+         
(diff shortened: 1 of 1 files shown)


docs/code_snippet.rst
```diff
@@ -58,7 +58,10 @@ will wait for 15 minutes each time it hits the rate limit.
                yield next(cursor)
            except tweepy.RateLimitError:
                time.sleep(15 * 60)
-   
+               return
```
dougharris, line 61: I don't think the return should be here. Since this is in a `while True` loop, just sleep for 15 minutes and then it tries the call to twitter again.

Harmon758:
This is stale now, with https://github.com/tweepy/tweepy/commit/fe5a0508f97326e0be4fddca5db65a655607c2f9 and https://github.com/tweepy/tweepy/commit/5fa6e73bbe677cf2527af7187012f665b825639d.

This example never really made much sense to include in the documentation anyway, since `wait_on_rate_limit` handles this.

PR #1653, opened 2021-09-01
tweak streaming docs to show correct parameter

jessesheidlower:
The actual parameter the programmer passes to the function is "languages", and takes a list; "language=en" is what's happening behind the scenes but that's not how you would code this.


Harmon758:
This documentation was taken from the Twitter API documentation, but this is stale now anyway, with the Twitter API [v1.1 statuses/sample endpoint retired](https://twittercommunity.com/t/deprecation-announcement-removing-compliance-messages-from-statuses-filter-and-retiring-statuses-sample-from-the-twitter-api-v1-1/170500) and the [v1.1 statuses/filter endpoint deprecated](https://twittercommunity.com/t/announcing-the-deprecation-of-v1-1-statuses-filter-endpoint/182960) now.

PR #1683, opened 2021-10-04
Corrected spelling mistake on API.search_tweets method

Syndorik:
Just corrected a small mistake on API.search_tweets method


Harmon758 (approved):
Nice catch! This seems to have been [a typo copied and pasted from Twitter's API documentation in 2009](https://web.archive.org/web/20090420040222/http://apiwiki.twitter.com/Twitter-Search-API-Method:-search) that was never updated when Twitter corrected it.

PR #1690, opened 2021-10-28
twitter_streamer_ex.py

gitelians:
That's a simple example which explain how to set your code in order to download real-time data from twitter, in according to the last tweepy version (4.1.0)


Harmon758:
Sorry, but there are a lot of issues with this, and it's too low quality as an example.
Superseded by https://github.com/tweepy/tweepy/commit/56f15473917155a44c29e49c97cba1ef4cef159d

PR #1733, opened 2021-12-17
Minor fix(es) to Client's _make_request

kanguyen-vn:
The current code in master has 

```
if "poll" in includes:
            includes["polls"] = [Poll(poll) for poll in includes["polls"]]
```

on lines 189-190 in `client.py`. Because the key in the `includes` dictionary is actually `polls` instead of `poll`, line 190 is never executed, and thus the poll dictionaries fetched using the Twitter get request are never converted to Poll objects. This affects code consistency because all of the dictionaries in `includes["tweets"]`,  `includes["users"]`, `includes["media"]`, and `includes["places"]` get transformed into Tweet, User, Media, and Place objects respectively.

The other 2 fixes are trivial style fixes only (added line break so the line isn't too long, and removed spaces surrounding = for styling consistency).

### DIFF
--- a/tweepy/client.py
+++ b/tweepy/client.py
@@ -156,7 +156,8 @@ class Client:
                 if param_value.tzinfo is not None:
                     param_value = param_value.astimezone(datetime.timezone.utc)
                 request_params[param_name] = param_value.strftime(
-                    "%Y-%m-%dT%H:%M:%S.%fZ")
+                    "%Y-%m-%dT%H:%M:%S.%fZ"
+                )
                 # TODO: Constant datetime format string?
             else:
                 request_params[param_name] = param_value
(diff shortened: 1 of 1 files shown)


tweepy/client.py
```diff
@@ -155,7 +155,8 @@ def _make_request(self, method, route, params={}, endpoint_parameters=None,
             elif isinstance(param_value, datetime.datetime):
                 if param_value.tzinfo is not None:
                     param_value = param_value.astimezone(datetime.timezone.utc)
-                request_params[param_name] = param_value.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+                request_params[param_name] = param_value.strftime(
+                    "%Y-%m-%dT%H:%M:%S.%fZ")
```
Harmon758, line 159:
```suggestion
                    "%Y-%m-%dT%H:%M:%S.%fZ"
                )
```

Harmon758 (approved):

Harmon758:
Good catches! Thanks for the PR.

PR #1746, opened 2021-12-27
Create value52.py

PvrpleBlvck:
'''Okay, so the API allows a maximum of 1000 items or 50 pages. That's where the issue is coming from. It's more like a limitation.
     You either have just work with what you have but if you wanna have fun coding

     I think maybe the best way around it would be to use a function
      ( that iterates through the first 1000 items, 
      and break out of the function) 
      run the function again but iterating from the 1000th item to the 2000th. And so forth
      Maybe try the python NEXT function'''

# Thank you. I would appreciate the feedback and if anyone has a better of solving this issue also welcome to get in touch with on the comment section

### DIFF
--- a/value52.py
+++ b/value52.py
@@ -1,0 +1,23 @@
+    '''Okay,so the API allows a maximum of 1000 items or 50 pages. That's where the issue is coming from. It's more like a limitation.
+     You either have just work with what you have but if you wanna have fun coding
+
+     I think maybe the best way around it would be to use a function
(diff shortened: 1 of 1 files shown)


Harmon758:
I'm not sure what the purpose of this pull request is, but this is not something that makes sense to include in the library, even as an example.

PvrpleBlvck:
Hie, apologies am new to this. I was trying to solve valueerror52. Wasn't
sure to really go about it. I was trying to update the issue

Issue #1790, opened 2022-01-25
Add support for COVID-19 stream


utanashati:
Hi! I've been successfully using tweepy for streaming from POST statuses/filter in a research lab. Now I'd like to rewrite the app that streams from the [COVID-19 stream](https://developer.twitter.com/en/docs/labs/covid19-stream/overview) with tweepy as well, since it has all the necessary wrappings for Twitter connections. The endpoint is [GET /labs/1/tweets/stream/covid19](https://developer.twitter.com/en/docs/labs/covid19-stream/api-reference/get-tweets-stream-covid19) + partition query (4 partitions total to boost throughput, this is a huge stream with 10mils tweets per day).

I imagine this could easily be done by adding a function like `tweepy.Stream.filter/sample` to `streaming.py` and allowing custom urls in `tweepy.Stream._connect`. I think adding this feature could help other researchers to stream from the COVID-19 stream easier.

Harmon758:
This was originally part of #1472 with support for Labs.
I asked one of the Twitter Developer Advocates I'm in touch with about adding support for it and inquired about the state of Labs when Twitter API v2 launched, as I believe this is the only remaining endpoint in Labs. From my understanding, it's in the unique position of being supported but [no longer accepting applications](https://twittercommunity.com/t/new-covid-19-stream-endpoint-available-in-twitter-developer-labs/135540/7). We discussed how this was the case and how anyone who had access likely already established other methods to use the endpoint, so at the time, I came to the decision of not working on adding support for it.

I'm not adverse to doing so, but there's also a question of how much longer this endpoint will continue to exist. I'm also not sure it would a good idea to advertise support of it, as that might lead to questions about how to obtain access to it. I've reached back out to Twitter about this, and I'm going to wait to hear back before proceeding with anything.

andypiper:
Hi from @TwitterDev - as mentioned above, we're delighted to see the interest in adding this endpoint as formally supported in tweepy, but since there are no plans to provide new access (and this endpoint will in the future be retired, as it was always intended as a point-in-time research tool), and we would not want to set an expectation that this functionality is available to new developers, it might be an idea to leave this as a separate branch rather than part of the core code. Just my 2c.

Harmon758:
[...]

PR #1791, opened 2022-01-25
Support for COVID-19 stream

utanashati:
#1790

Added a method `tweepy.Stream.covid` to stream from [Twitter COVID-19 stream](https://developer.twitter.com/en/docs/labs/covid19-stream/overview). For this, I had to put exact endpoint URLs to `filter` and `sample` instead of endpoint names to be able to use `_connect` with a different type of URL: `https://api.twitter.com/labs/1/tweets/stream/covid19`.

### DIFF
--- a/tweepy/streaming.py
+++ b/tweepy/streaming.py
@@ -19,6 +19,7 @@ import urllib3
 import tweepy
 from tweepy.errors import TweepyException
 from tweepy.models import Status
+from tweepy.auth import OAuth2AppHandler
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


utanashati:
Fixed an oauth issue. Will also modify the async streaming accordingly.

Harmon758 (changes requested):
As part of adding support for streaming with Twitter API v2 with https://github.com/tweepy/tweepy/commit/86244c1a82a1852d04f3695b03201363f5d5eafd, I've refactored `Stream` to inherit from a new `BaseStream` class. It should be easier now to create a class for COVID-19 streaming that inherits from `BaseStream`. That would improve on this interface, as `Stream` should be specific to streaming with Twitter API v1.1.

Harmon758:
With the COVID-19 stream endpoint being deprecated, as part of [Twitter Developer Labs being deprecated](https://developer.twitter.com/en/blog/product-news/2022/deprecating-twitter-developer-labs), and set to be retired on October 31, 2022, I'm going to close this now.

PR #1850, opened 2022-03-25
Manual Refresh Token feature

lackingdev:
This adds a refresh token feature for OAuth2 user tokens. The one important thing is that either the same scopes need to be passed in oauth2_user_handler or scopes need to be set to None.


mintyPT:
Used this and worked flawlessly. What needs to be done to merge it?

PR #1910, opened 2022-06-09
Add missing "variants" to Media Object

New-dev0:
Media object was missing the variants attribute/data.
even though it was present in https://developer.twitter.com/en/docs/twitter-api/data-dictionary/object-model/media

### DIFF
--- a/tweepy/media.py
+++ b/tweepy/media.py
@@ -60,7 +60,7 @@ class Media(DataMapping):
         A description of an image to enable and support accessibility. Can be
         up to 1000 characters long. Alt text can only be added to images at the
         moment.
-    variants: List[Dict] | None
+    variants: list[dict] | None
         Each media object may have multiple display or playback variants,
         with different resolutions or formats
 


tweepy/media.py
```diff
@@ -60,6 +60,9 @@ class Media(DataMapping):
         A description of an image to enable and support accessibility. Can be
         up to 1000 characters long. Alt text can only be added to images at the
         moment.
+    variants: List[Dict] | None
```
Harmon758, line 63:
```suggestion
    variants: list[dict] | None
```

Harmon758 (changes requested):
Thanks for the PR.

This was [previously undocumented](https://web.archive.org/web/20220516192303/https://developer.twitter.com/en/docs/twitter-api/data-dictionary/object-model/media) and seems to have been added without any announcement or notice.

A version changed directive needs to be added to the documentation.
This should end up being released with v4.12.

Harmon758 (approved):

PR #1921, opened 2022-07-09
Get Context Entities from Twitter Dev Repo

s2t2:
The Twitter API [docs](https://developer.twitter.com/en/docs/twitter-api/annotations/overview) reference a list of context entity records kept in a [github repo](https://github.com/twitterdev/twitter-context-annotations). This PR adds a method to the `Client` to fetch this data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@ env.sh
 .pdbrc
 tweepy.egg-info
 
+.vscode/
+
 # Created by https://www.gitignore.io/api/vim,python
 
 ### Python ###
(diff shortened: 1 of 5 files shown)


freddie71010:
Should we really be adding another package (`pandas`) to the setup just for one function? Is this functionality that important? Seems like that's the start of bloatware :\

s2t2:
I understand the concern.

Although pandas is pretty standard for many python projects, perhaps we
could add the dependency as an optional "extra".

Issue #1871, opened 2022-04-22
wait_on_rate_limit when using Paginator with v2 search_all_tweets endpoint


mirandrom:
The following code will immediately output a Rate limit exceeded error and sleep for 900s.

```
import tweepy
import os

client = tweepy.Client(bearer_token=os.environ["TWITTER_BEARER_TOKEN"], wait_on_rate_limit=True)

query = "tweepy"

for response in tweepy.Paginator(client.search_all_tweets, query=query):
  process_response(response)
```

I think what happens here is due to the following rate limits for `GET /2/tweets/search/all`:

300 requests / 15 minsPER APP
1 requests / secondPER USER
1 requests / secondPER APP

I think this might be due to the following line, where the sleep time is taken from the response headers:
https://github.com/tweepy/tweepy/blob/0eac99beedf7f76c9587d91742453312dbca13d8/tweepy/client.py#

I found that changing my code to the below significantly increased the throughput when using the paginator.
Would it make sense to update the waiting strategy in the client to try and sleep for 1s before using the response header value?

```
from time import sleep
import tweepy
import os

client = tweepy.Client(bearer_token=os.environ["TWITTER_BEARER_TOKEN"], wait_on_rate_limit=True)

query = "tweepy"

for response in tweepy.Paginator(client.search_all_tweets, query=query):
  sleep(1)
  process_response(response)
```

Harmon758:
This is a duplicate of #1688, and there's an [FAQ section in the documentation about this](https://tweepy.readthedocs.io/en/v4.8.0/faq.html#why-am-i-getting-rate-limited-so-quickly-when-using-client-search-all-tweets-with-paginator).

I'm open to reconsidering how Tweepy handles this though.

mirandrom:
Ah wonderful thank you, I had missed those with the keywords I searched before posting this. Apologies for the duplicate.
I think it could be helpful  to have a note in the Pagination docs. I completely missed the FAQ; and it's easy to not even notice the issue if running jobs without monitoring them.

mirandrom:
Happy to take a stab at it if its something you decide is worthwhile.

Issue #1688, opened 2021-10-14
Add a delay before search_all_tweets calls


jdfoote:
`search_all_tweets` has a 1 call per second limit. Currently, `tweepy` quickly makes a few calls, receives a `429`, and then waits for nearly 15 minutes. By simply adding a sleep of 1 second per call, you can make 300 calls in that 15 minutes.

It seems like the place to put this may be in the `pagination.py` file, as the delay is only a problem when making multiple calls? I'd be happy to make a one-line PR, but I'm not sure where the devs would want to put the sleep.

Harmon758:
I'm aware of this issue, but I haven't determined the best way to resolve it yet.
Ideally, I'd rather not have to hard-code the rate limit into the library, but that might end up being necessary, as this rate limit isn't in the response headers and seems to only be declared in the endpoint's documentation.
For now, a workaround is simply to sleep 1 second between each request, while iterating through the `Paginator` responses.

Harmon758:
Upon consideration, it might not be best to handle this within Tweepy.

There might be users who do processing within the loop that takes up a significant amount of time or even takes longer than a second, so a simple 1 second sleep wouldn't be ideal.

The alternative would be saving the timestamp that the last request was made at and sleeping until a second has passed, but that might end up being exactly a second and there would need to be considerations for adding some jitter.

I think you're right and the simplest way forward right now would be to document it and allow the user to handle it themselves.
If it becomes necessary at some point, handling of this rate limit can always be added later.

<details>
  <summary>PS</summary>

  @jdfoote I saw your video 👍

  Some notes:
  - `start_time` and `end_time` can be `datetime` objects if you don't want to have to remember the string format that Twitter's API requires
  - Tweepy's API v2 models have a `data` attribute (still needs to be documented, but it's not going anywhere) that provides the entire data dictionary
</details>

xrisk:
@Harmon758  This rate limit should be present in at least the `flatten` api call I think. There is no way to insert 1 second sleeps there. On a related note, I feel this should be documented on the main docs page for pagination, not in the FAQ.

PR #1923, opened 2022-07-17
Optimize request intervals for full archive search with OAuth 2.0 Bearer Token for pagination.py

TomatenMarc:
With this PR I would like to solve the issues raised in #1688 #1907 and #1871 of missing wait time between requests for Twitter api v2, which causes a direct exceeding of rate limits during a full archive search.

The problem seems not to be in [https://github.com/tweepy/tweepy/blob/master/tweepy/client.py](url) as in #1871, but rather in [https://github.com/tweepy/tweepy/blob/master/tweepy/pagination.py](url) as described in #1688.

It is a way to simply wait one second at the end of each request and the processing time of the data, but this always requires consideration and knowledge of the problem by the user.
Likewise, it may happen that the optimal time windows of 1 request per second and 300 requests per 900 requests (3 seconds for request + processing) cannot be met if 1 second is always added to the request and processing time.

Mainly, though, I think the problem should be solved within Tweepy, since the user (at least I did and take some time to find the problem) assumes that Tweepy implements Twitter's guidelines.

Since the Twitter guidelines only require a limit of 1 request per second for the full archive search `/2/tweets/search/all` in combination with OAuth 2.0 Bearer Token cf. https://developer.twitter.com/en/docs/twitter-api/tweets/search/migrate it seems appropriate to measure the time from the beginning in `__next__` of [https://github.com/tweepy/tweepy/blob/master/tweepy/pagination.py](url) and to fill up the next second after receiving the response.

Cheers :-)

### DIFF
--- a/tweepy/pagination.py
+++ b/tweepy/pagination.py
@@ -1,7 +1,7 @@
 # Tweepy
 # Copyright 2009-2022 Joshua Roesslein
 # See LICENSE for details.
-
+import time
 from math import inf
 
 
@@ -78,6 +78,8 @@ class PaginationIterator:
         return self
 
     def __next__(self):
+        t0: float = time.time()
+
         if self.reverse:
             pagination_token = self.previous_token
         else:
@@ -88,8 +90,9 @@ class PaginationIterator:
 
         # https://twittercommunity.com/t/why-does-timeline-use-pagination-token-while-search-uses-next-token/150963
         if self.method.__name__ in (
-            "search_all_tweets", "search_recent_tweets",
-            "get_all_tweets_count"
+                "search_all_tweets",
+                "search_recent_tweets",
+                "get_all_tweets_count"
         ):
             self.kwargs["next_token"] = pagination_token
         else:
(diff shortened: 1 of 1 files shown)


vgewilliam:
> With this PR I would like to solve the issues raised in #1688 #1907 and #1871 of missing wait time between requests for Twitter api v2, which causes a direct exceeding of rate limits during a full archive search.
> 
> The problem seems not to be in [https://github.com/tweepy/tweepy/blob/master/tweepy/client.py](url) as in #1871, but rather in [https://github.com/tweepy/tweepy/blob/master/tweepy/pagination.py](url) as described in #1688.
> 
> It is a way to simply wait one second at the end of each request and the processing time of the data, but this always requires consideration and knowledge of the problem by the user. Likewise, it may happen that the optimal time windows of 1 request per second and 300 requests per 900 requests (3 seconds for request + processing) cannot be met if 1 second is always added to the request and processing time.
> 
> Mainly, though, I think the problem should be solved within Tweepy, since the user (at least I did and take some time to find the problem) assumes that Tweepy implements Twitter's guidelines.
> 
> Since the Twitter guidelines only require a limit of 1 request per second for the full archive search `/2/tweets/search/all` in combination with OAuth 2.0 Bearer Token cf. https://developer.twitter.com/en/docs/twitter-api/tweets/search/migrate it seems appropriate to measure the time from the beginning in `__next__` of [https://github.com/tweepy/tweepy/blob/master/tweepy/pagination.py](url) and to fill up the next second after receiving the response.
> 
> Cheers :-)

Hi dude, thanks for your contribution, it works for me.

Riahiamirreza (approved):

TomatenMarc:
@vgewilliam Is there anything else to take into account or can this be merged? :-)

PR #1861, opened 2022-04-13
Fix `AttributeError` by accessing `meta` field when `return_type` is dict

lqhuang:
I meet totally the same error to #1843 

```
Traceback (most recent call last):
  File "${MY_PY_ENV}/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "${MY_PY_ENV}/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "${MY_WORKSPACE}/twitter_fetcher/fetch_tweet_timeline.py", line 174, in <module>
    main()
  File "${MY_WORKSPACE}/twitter_fetcher/fetch_tweet_timeline.py", line 84, in main
    for i, resp in enumerate(
  File "${MY_PY_ENV}/lib/python3.10/site-packages/tweepy/pagination.py", line 100, in __next__
    self.previous_token = response.meta.get("previous_token")
AttributeError: 'dict' object has no attribute 'meta'
```

My solution is to add an `isinstance` condition, it's very simple but useful. Any improvement or suggest for this PR is welcome.

Regards.

PR #1985, opened 2022-10-14
pagination.py fixed

calzateu:
The error "AttributeError: 'Response' object has no attribute 'meta'" was corrected. This happens when I initialize the Client object with the "return_type=requests.Response" parameter.


Harmon758:
This is a duplicate of #1861, doesn't handle the dictionary return type, and would raise a `NameError` with `Response` not being defined/imported.

Issue #1989, opened 2022-10-23
Include JSON response in V2 requests


jdfoote:
The V1 requests include the raw JSON response, which I like to save as backup data. Would it be possible to have it included in the V2 responses as well? E.g., `search_all_tweets`?

Harmon758:
Note, setting `return_type` to `dict` for `Client` or `AsyncClient` will have their methods return the raw JSON as a dictionary.

I'm assuming that this is a feature request to include the raw JSON data in `Response` though, as that would allow Tweepy models to be used, while still having the raw JSON data available to save as backup data.

I'm considering changing `Response` based on this and other use cases, like #1984.
It would almost certainly be a breaking change though, so it would have to wait until a v5 release.

For now, as a workaround, with #1997, which resulted from [a conversation about this in the Tweepy Discord server](https://discord.com/channels/432685901596852224/1035394230434152448), you can use a `requests.Response` or `dict` `return_type` to access the JSON, headers, etc. before passing the JSON as a dictionary to `Client._construct_response` or `AsyncClient._construct_response`. Note, you'll also need to pass the `data_type` for any models in `data` if you want them to be Tweepy models.

jdfoote:
Thanks so much for the response (and for all of the work on `tweepy`!)

You have understood the request perfectly - I'd love to have the JSON included in the `Response` object.

This workaround seems great for now. Thanks so much!

PR #1997, opened 2022-10-28
Breakout Response creation into separate method

JEnduriumK:
As a workaround to folks who would like the headers from a `requests.Response` but also the nice, neat `tweepy.Response` data type? I think #1984 and #1989 might be related?


JEnduriumK:
I did no testing of this code *because I don't know how*. I'm new to Python.

Harmon758 (approved):

PR #2009, opened 2022-11-11
Add fields documentation for Direct Message events

qin-yu:
Hi, I recently used the newly added `Client.get_direct_message_events()` and found the documentation missing. Since I read the Twitter doc carefully, I'll just update the Tweepy doc. It's the first time that I edited a `.rst` Sphinx doc file. Hopefully it's not too wrong. Three changes in two files:

- `docs\expansions_and_fields.rst`: 
  - `expansions` section modified
  - `dm_event_fields` section created
- `tweepy\client.py`: 
  - `get_direct_message_events`'s link to doc added

### DIFF
--- a/docs/expansions_and_fields.rst
+++ b/docs/expansions_and_fields.rst
@@ -8,13 +8,13 @@ Expansions and Fields
 ``expansions``
 --------------
 `Expansions`_ enable you to request additional data objects that relate to the
-originally returned List, Space, Tweets, Users, or DirectMessageEvents. Submit
-a list of desired expansions in a comma-separated list without spaces. The ID
-that represents the expanded data object will be included directly in the List,
-Space, Tweet, User, or DirectMessageEvents data object, but the expanded object
-metadata will be returned within the ``includes`` response object, and will
-also include the ID so that you can match this data object to the original
-Space, Tweet, User or DirectMessageEvents object.
+originally returned List, Space, Tweets, Users, or Direct Message conversation
+events. Submit a list of desired expansions in a comma-separated list without
+spaces. The ID(s) that represent(s) the expanded data object(s) will be
+included directly in the List, Space, Tweet, user, or event data object, but
+the expanded object metadata will be returned within the ``includes`` response
+object, and will also include the ID so that you can match this data object to
+the original Space, Tweet, User or Direct Message conversation event object.
 
 At this time, the only expansion available to endpoints that primarily return
 List objects is ``expansions=owner_id``. You will find the expanded user data
--- a/docs/expansions_and_fields.rst
+++ b/docs/expansions_and_fields.rst
@@ -41,18 +41,13 @@ using this parameter:
 * Attached place’s object
 * Any referenced Tweets’ object
 
-For methods that returns DirectMessageEvents, the following data objects can be
-expanded using this parameter:
+For methods that returns Direct Message conversation events, the following data
+objects can be expanded using this parameter:
 
-* ``sender_id`` - The ID of the account that sent the message, or who invited
-a participant to a group conversation
-* ``partricipants_ids`` - An array of account IDs. For ParticipantsJoin and
-ParticipantsLeave events this array will contain a single ID of the account
-that created the event
-* ``attachments`` - Provides media IDs for content that has been uploaded to
-Twitter by the sender
-* ``referenced_tweets`` - If a Tweet URL is found in the text field, the ID of
-that Tweet is included in the response
+* The user object for the message sender.
+* Attached media's object.
+* Any referenced Tweet's object.
+* The user object for who is joining or leaving group conversations.
 
 At this time, the only expansion available to endpoints that primarily return
 user objects is ``expansions=pinned_tweet_id``. You will find the expanded
--- a/docs/expansions_and_fields.rst
+++ b/docs/expansions_and_fields.rst
@@ -157,38 +157,6 @@ Specify the desired fields in a comma-separated list without spaces between
 commas and fields. These specified user fields will display directly in the
 user data objects.
 
-.. _dm_event_fields_parameter:
-
-``dm_event_fields``
----------------
-
-For methods that return DirectMessageEvents, this `fields`_ parameter enables
-you to select which specific `Direct Message event fields`_ will deliver in
-each returned DirectMessageEvent. Specify the desired fields in a comma-
-separated list without spaces between commas and fields. When requesting DM
-events, there are three default event object attributes, or fields, included:
-``id``, ``event_type``, and ``text``. To receive additional event fields, use
-the fields parameter ``dm_event_fields`` to select others. Other available
-event fields include the following: ``dm_conversation_id``, ``created_at``,
-``sender_id``, ``attachments``, ``participant_ids``, and ``referenced_tweets``.
-
-Several of these fields provide the IDs of other Twitter objects related to the
-Direct Message event:
-
-* ``sender_id`` - The ID of the account that sent the message, or who invited
-a participant to a group conversation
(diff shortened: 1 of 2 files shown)


docs/expansions_and_fields.rst
```diff
@@ -8,12 +8,13 @@ Expansions and Fields
...
-user data object, but the expanded object metadata will be returned within the
-``includes`` response object, and will also include the ID so that you can
-match this data object to the original Space, Tweet, or user object.
+originally returned List, Space, Tweets, Users, or DirectMessageEvents. Submit
+a list of desired expansions in a comma-separated list without spaces. The ID
+that represents the expanded data object will be included directly in the List,
+Space, Tweet, User, or DirectMessageEvents data object, but the expanded object
+metadata will be returned within the ``includes`` response object, and will
+also include the ID so that you can match this data object to the original
+Space, Tweet, User or DirectMessageEvents object.
```
Harmon758, line 17: This section is taken from the description for the expansions parameter for endpoints that return each type of object, e.g. https://developer.twitter.com/en/docs/twitter-api/direct-messages/lookup/api-reference/get-dm_conversations-with-participant_id-dm_events for DM events.

```suggestion
originally returned List, Space, Tweets, Users, or Direct Message conversation
events. Submit a list of desired expansions in a comma-separated list without
spaces. The ID(s) that represent(s) the expanded data object(s) will be
included directly in the List, Space, Tweet, user, or event data object, but
the expanded object metadata will be returned within the ``includes`` response
object, and will also include the ID so that you can match this data object to
the original Space, Tweet, User or Direct Message conversation event object.
```

docs/expansions_and_fields.rst
```diff
@@ -40,6 +41,19 @@ using this parameter:
...
+
+* ``sender_id`` - The ID of the account that sent the message, or who invited
+a participant to a group conversation
+* ``partricipants_ids`` - An array of account IDs. For ParticipantsJoin and
+ParticipantsLeave events this array will contain a single ID of the account
+that created the event
+* ``attachments`` - Provides media IDs for content that has been uploaded to
+Twitter by the sender
+* ``referenced_tweets`` - If a Tweet URL is found in the text field, the ID of
+that Tweet is included in the response
```
Harmon758, line 55: I have plans to improve this section of the documentation in the future to include information like this from https://developer.twitter.com/en/docs/twitter-api/expansions, but for now, for consistency, this section should also be taken from the description for the expansions parameter for endpoints that return the object type, e.g. https://developer.twitter.com/en/docs/twitter-api/direct-messages/lookup/api-reference/get-dm_conversations-with-participant_id-dm_events for DM events.

```suggestion
For methods that returns Direct Message conversation events, the following data
objects can be expanded using this parameter:

* The user object for the message sender.
* Attached media's object.
* Any referenced Tweet's object.
* The user object for who is joining or leaving group conversations.
```

This should probably also be after the following section on user objects.

docs/expansions_and_fields.rst
```diff
@@ -148,6 +162,38 @@ Specify the desired fields in a comma-separated list without spaces between
...
+
+The ``sender_id``, ``participant_ids``, ``referenced_tweets.id``, and
+``attachments.media_keys`` `expansions`_ are available to expand on these Twitter
+object IDs.
+
```
Harmon758, line 196: Similarly, the sections like this for other object types are taken from the description for the respective fields parameter for endpoints that return each type of object, e.g. https://developer.twitter.com/en/docs/twitter-api/direct-messages/lookup/api-reference/get-dm_conversations-with-participant_id-dm_events for DM events.

However, you can see that the documentation for `dm_event.fields` is different, so this section is unnecessary. Again, I have plans to improve this documentation as a whole to include documentation from https://developer.twitter.com/en/docs/twitter-api/expansions, so for now, for consistency, this section shouldn't be here.

Harmon758 (changes requested):
A lot of documentation was missing when the v2 DM endpoints were launched. I was waiting until more of it was added before fully updating all the relevant documentation for Tweepy. A lot of it has been added now, in the past few days, so I'll be updating other documentation relating to DMs soon.

docs/expansions_and_fields.rst
qin-yu: Oh I see, the term "___ object" is used in a natural language context, and Twitter's official documents do use lower cases for "user" and "event". Also, indeed, query parameters' descriptions in all three Direct Messages lookup API references ([1](https://developer.twitter.com/en/docs/twitter-api/direct-messages/lookup/api-reference/get-dm_events) [2](https://developer.twitter.com/en/docs/twitter-api/direct-messages/lookup/api-reference/get-dm_conversations-with-participant_id-dm_events) [3](https://developer.twitter.com/en/docs/twitter-api/direct-messages/lookup/api-reference/get-dm_conversations-dm_conversation_id-dm_events)) use the term "Direct Message conversation event", though in the informal abstract/intro sections they use "Direct Messages (DM) events", "Direct Messages" and "Messages" interchangeably. 

Very accurate review! I hope this PR didn't bring more trouble than convenience...

docs/expansions_and_fields.rst
qin-yu: I see. I adapted the description here https://developer.twitter.com/en/docs/twitter-api/data-dictionary/object-model/dm-events to make this section in our doc. If it is not necessary then l'll remove line 165-195.

Many thanks!

qin-yu: I'll remove this section, but if we want some completeness before the next big doc improvement, we can put the following code back later just in case people thought the doc is not up-to-date:

```rst
.. _dm_event_fields_parameter:

``dm_event_fields``
---------------

Extra fields to include in the event payload. `id`, `text` and `event_type` are
returned by default.

.. _Direct Message event fields: https://developer.twitter.com/en/docs/twitter-api/data-dictionary/object-model/dm-events
```

Harmon758 (approved):

docs/expansions_and_fields.rst
Harmon758: > Oh I got it. You meant the adapted intro from https://developer.twitter.com/en/docs/twitter-api/data-dictionary/object-model/dm-events is not necessary, and, for now, I should use the _description_ for `dm_event.fields`. However, there is only one endpoint using `dm_event.fields` so I don't need to complete this documentation?

Correct, I've added that documentation to `DirectMessageEvent` itself now, with https://github.com/tweepy/tweepy/commit/c689548cb1b4b9c2f69249d67534a735b931b9b5.

docs/expansions_and_fields.rst
Harmon758: I've added the tables from https://developer.twitter.com/en/docs/twitter-api/expansions now, with https://github.com/tweepy/tweepy/commit/7df645328bdf26e5c4c569392d5d2252b1527477.

docs/expansions_and_fields.rst
Harmon758: > I'll remove this section, but if we want some completeness in this doc page before the next big doc improvement (and also for the consistency in Client methods' docstrings), we can put the following code back later just in case people thought the doc is not up-to-date:
> 
> ```
> .. _dm_event_fields_parameter:
> 
> ``dm_event_fields``
> ---------------
> 
> Extra fields to include in the event payload. `id`, `text` and `event_type` are
> returned by default.
> 
> .. _Direct Message event fields: https://developer.twitter.com/en/docs/twitter-api/data-dictionary/object-model/dm-events
> ```

For the sake of completeness, I've added a section for this now, with https://github.com/tweepy/tweepy/commit/4b0fa90e91eb2b67dfd33f0d27b148e95ea05f65.
I don't think it's necessary to replace or add it to the documentation for the `Client` and `AsyncClient` methods that have this parameter yet though.

PR #2017, opened 2022-11-15
Add media/subtitles/create function and test

jmassa23:


jmassa23:
That was an accident! sorry!

PR #2033, opened 2023-01-14
Add types to `auth.py`

Viicos:
Noticed there's no code formatting applied to this library. We might want to use `isort` and a code formatter (e.g `black`), as type hints addition will require a lot of imports.

### DIFF
--- a/tweepy/auth.py
+++ b/tweepy/auth.py
@@ -1,9 +1,11 @@
 # Tweepy
 # Copyright 2009-2022 Joshua Roesslein
 # See LICENSE for details.
+from __future__ import annotations
 
 import logging
 import warnings
+from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


tweepy/auth.py
```diff
@@ -117,8 +124,8 @@ class OAuthHandler(OAuth1UserHandler):
         Use :class:`OAuth1UserHandler` instead.
     """
 
-    def __init__(self, consumer_key, consumer_secret, access_token=None,
```
Viicos, line 120: This is probably not that important, but maybe we could use a better a way of deprecating old name classes here. See [this comment](https://stackoverflow.com/questions/9008444/how-to-warn-about-class-name-deprecation#comment80382529_26130976) on why it could be problematic, and a few options given in the SO thread

Issue #2037, opened 2023-01-19
Pagination v2: flatten breaks when return_type=requests.Response


rhymes:
Currently, if `tweepy.Client`'s `return_type` is set to `requests.Response` and `tweepy.Paginator` is used, the client fails with `.flatten()`:

```
>               raise RuntimeError(
                    f"Paginator.flatten does not support the {type(response)} "
                    f"return type for {self.method.__qualname__}"
                )
E               RuntimeError: Paginator.flatten does not support the <class 'requests.models.Response'> return type for Client.get_blocked
```

That is because, even though `PaginationIterator` supports it:

https://github.com/tweepy/tweepy/blob/0cd96b1918e5e920eb9f8fe4ba303ab5ec899c65/tweepy/pagination.py#L132-L133

the  `flatten` method doesn't:

https://github.com/tweepy/tweepy/blob/0cd96b1918e5e920eb9f8fe4ba303ab5ec899c65/tweepy/pagination.py#L70-L75

I'm going to open a PR soon

rhymes:
Hi @Harmon758, thanks for the reply, I think I understand what you mean. If the data type to flatten is either a dictionary or a `tweepy.Response` then flatten can extract the data, a `requests` response is just a response (though it still contains the data...).

The issue I'm facing is that I'm using `return_type=requests.Response` because I want to examine the headers sent by Twitter, but if the business logic calls `tweepy.Paginator(client.get_something).flatten()` on a `client` object, set to `return_type=requests.Response` then the code breaks.

What do you suggest?

Harmon758:
You can either process the `requests.Response` yourself, make two separate requests, or use the workaround outlined in https://github.com/tweepy/tweepy/issues/1989#issuecomment-1294528006 to construct the `tweepy.Response` and iterate over the `data` yourself.

Note, this was initially presented as an XY problem, which you should try to avoid doing in the future.

rhymes:
Thank you!

PR #2038, opened 2023-01-19
Fix flatten for sync and async paginators

rhymes:
This closes https://github.com/tweepy/tweepy/issues/2037

I added the code for the synchronous and asynchronous paginators.

I've added no tests as there aren't any...

Thank you!


Harmon758:
This behavior does not make sense, as it would end up returning dictionaries when the `return_type` is set to `requests.Response` and/or `aiohttp.ClientResponse`.

This also ends up being equivalent to the existing supported option of setting the `return_type` to `dict` anyway, which is behavior that does make sense.

PR #2046, opened 2023-01-25
Add field 'verified_type' to User

WagnerChristoph:
I noticed in the the [Twitter Api Changelog](https://developer.twitter.com/en/updates/changelog) that a new field 'verified_type' has been added to the user model.

### DIFF
--- a/tweepy/user.py
+++ b/tweepy/user.py
@@ -63,6 +63,8 @@ class User(HashableID, DataMapping):
         The URL specified in the user's profile, if present.
     verified : bool | None
         Indicates if this user is a verified Twitter User.
+    verified_type : "blue" | "business" | "government" | "none"
+        Indicates the type of verification for the Twitter account.
     withheld : dict | None
         Contains withholding details for `withheld content`_, if applicable.
 
@@ -77,7 +79,7 @@ class User(HashableID, DataMapping):
     __slots__ = (
         "data", "id", "name", "username", "created_at", "description",
         "entities", "location", "pinned_tweet_id", "profile_image_url",
-        "protected", "public_metrics", "url", "verified", "withheld"
+        "protected", "public_metrics", "url", "verified", "verified_type", "withheld"
     )
 
     def __init__(self, data):
@@ -103,6 +105,7 @@ class User(HashableID, DataMapping):
         self.public_metrics = data.get("public_metrics")
         self.url = data.get("url")
         self.verified = data.get("verified")
+        self.verified = data.get("verified_type")
         self.withheld = data.get("withheld")
 
     def __repr__(self):


tweepy/user.py
```diff
@@ -63,6 +63,8 @@ class User(HashableID, DataMapping):
         The URL specified in the user's profile, if present.
     verified : bool | None
         Indicates if this user is a verified Twitter User.
+    verified_type : "blue" | "business" | "government" | "none"
```
Harmon758, line 66: This should be the type, not the literal strings it could be, as there could easily be new and/or undocumented verified types. This could also be `None` when the field isn't in the data.

```suggestion
    verified_type : str | None
```

tweepy/user.py
```diff
@@ -63,6 +63,8 @@ class User(HashableID, DataMapping):
...
+        Indicates the type of verification for the Twitter account.
```
Harmon758, line 67: Since this field isn't documented for the User object yet, I think it'd be best to copy from the changelog:
```suggestion
        Indicates the type of verification a user account has (blue, business,
        government or none).
```

tweepy/user.py
```diff
@@ -77,7 +79,7 @@ class User(HashableID, DataMapping):
     __slots__ = (
         "data", "id", "name", "username", "created_at", "description",
         "entities", "location", "pinned_tweet_id", "profile_image_url",
-        "protected", "public_metrics", "url", "verified", "withheld"
+        "protected", "public_metrics", "url", "verified", "verified_type", "withheld"
```
Harmon758, line 82:
```suggestion
        "protected", "public_metrics", "url", "verified", "verified_type",
        "withheld"
```

tweepy/user.py
```diff
@@ -103,6 +105,7 @@ def __init__(self, data):
         self.public_metrics = data.get("public_metrics")
         self.url = data.get("url")
         self.verified = data.get("verified")
+        self.verified = data.get("verified_type")
```
Harmon758, line 108: This would override the existing `verified` attribute/field and doesn't match the documented attribute, field, or slots name.
```suggestion
        self.verified_type = data.get("verified_type")
```

Harmon758 (changes requested):
This isn't properly documented for the User object yet on https://developer.twitter.com/en/docs/twitter-api/data-dictionary/object-model/user, and the only announcement I saw about it was through https://twitter.com/suhemparack/status/1611085481395224576, but since, as you noted, it's on the changelog now, this should be fine to add.

A version changed directive also need to be added for this.
This will be released with v4.13.

Harmon758:
I've done this now with https://github.com/tweepy/tweepy/commit/9f7afaefa680da6a4f7864db99e12dd72a51cafa.

PR #2050, opened 2023-01-27
Fix KeyError in chunked upload method

frankdilo:
The Twitter API has stopped reliably returning the 'check_after_secs' key-value pair in their response to the chunked media upload endpoint.

This PR fixes the issue.

Until this is merged, we plan to start using my own forked branch.

CleanShot 2023-01-27 at 10 35 49@2x

### DIFF
--- a/tweepy/api.py
+++ b/tweepy/api.py
@@ -3685,7 +3685,7 @@ class API:
 
         if wait_for_async_finalize and hasattr(media, 'processing_info'):
             while media.processing_info['state'] in ('pending', 'in_progress'):
-                time.sleep(media.processing_info['check_after_secs'])
+                time.sleep(media.processing_info.get('check_after_secs', 5))
                 media = self.get_media_upload_status(media.media_id, **kwargs)
 
         return media


frankdilo:
This fix does solve the root cause and let's the upload process hang indefinitely if Twitter does not process the chunked upload correctly.

frankdilo:
https://twitter.com/typefully/status/1620377182005755904?s=61&t=y80msBCld3qwJ_QZsMIamw

Harmon758:
This is likely due to a bug with Twitter's API where the `processing_info` fields's `state` field gets stuck as `in_progress` rather than transitioning to `failed` when the upload errors, as it should based on [the documentation for the GET media/upload (STATUS) endpoint](https://developer.twitter.com/en/docs/twitter-api/v1/media/upload-media/api-reference/get-media-upload-status). This is what's happening in the image in the Tweet that you've linked.

I was made aware of this bug from [a support post in the Tweepy Discord server](https://discord.com/channels/432685901596852224/1060734723795730522), and I had a fix that handles this stashed that I've committed now, with https://github.com/tweepy/tweepy/commit/a68ce8658ab0908227dc486e73d02060e21174f8.

If this isn't the case for your original issue, then you'll need to provide more information, such as the full traceback and the relevant code or a reproducible example (MCVE/SSCCE). This PR would result in an infinite loop for the aforementioned bug, sleeping 5 seconds at a time, which also seems pretty arbitrary.

Issue #1606, opened 2021-06-07
Handle connection errors for requests made from `API` methods


Harmon758:
This includes:
- `ConnectionResetError: [Errno 54] Connection reset by peer` or `ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host` leading to `urllib3.exceptions.ProtocolError`, `requests.exceptions.ConnectionError` or `requests.exceptions.ChunkedEncodingError`, and `tweepy.error.TweepError`
- `ConnectionAbortedError: [WinError 10053] An established connection was aborted by the software in your host machine` leading to `requests.packages.urllib3.exceptions.ProtocolError`, `requests.exceptions.ConnectionError`, and `tweepy.error.TweepError`
- other types of `requests.exceptions.ConnectionError`, including `('Connection aborted.', OSError(0, 'Error'))` and `('Connection aborted.', BadStatusLine("''",))`
- `tweepy.error.TweepError: Failed to send request: HTTPSConnectionPool(host='api.twitter.com', port=443): Read timed out. (read timeout=60)`

in Tweepy v3.10.

This issue consolidates #617, #641, #645, #835, #937, #1205, #1528, #1555, #1559, and #1583.

johnkw:
See also bug #1219 where it was proposed consolidating those other bugs, along with a user-code callback function to determine when to stop retrying.

RyanAquino:
Any updates?

karjanme:
I experienced this error today in my app:
`HTTPSConnectionPool(host='api.twitter.com', port=443): Max retries exceeded with url: [. . .] (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at [. . .]>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution'))`

using Tweepy 3.8

PR #2073, opened 2023-02-24
Using `backoff.on_exception` decorator to solve #1606

ranggakd:
This PR consist as follows:
- [x] Added CoC (might suggest another yellow dotted bullet list below if the members don't mind)
- [x] Using `backoff.on_exception` decorator to #1606 (worked on my case, would like to take a look)

1. Yellow dotted bullet list, I just completed one at a time to see if you guys open with this kind of PR
yellow dotted

2. I came across this problem on Discord while trying to handle [another issue](https://discord.com/channels/432685901596852224/1078206406349557850). I noticed when I worked with my code to scrap certain tweets my connection always aborted since it took me so long to code and always rerun certain cells on my JupyterNB. So I chat with my ChatGPT, Gept, to solve this all together. I stumbled upon this #1606 while searching for the answer as well.

3. Gept was leading me to use the `backoff` module since I thought the connection need to be tried until the rate limit reset later.

> Initial brainstorming
initial brainstorming
> Initial suspect of the main problem while looking at the `Tweepy` source code
main problem
> Interested to other modules as solution
other modules

4. I tested it successfully with this snippet line of code. I was using `Client.get_liked_tweets` method and custom `Paginator` subclass to achieve this result below.
```python
import tweepy as tw
from backoff import on_exception, random_jitter, expo
from urllib3.exceptions import ProtocolError
from requests.exceptions import ConnectionError, ChunkedEncodingError
from tweepy.errors import TweepyException

class NewClient(tw.Client):
    @on_exception(expo, (ProtocolError, ConnectionError, ChunkedEncodingError, TweepyException), jitter=random_jitter, factor=2)
    def request(self, method, route, params=None, json=None, user_auth=False):
        return super().request(method, route, params, json, user_auth)
```

> Before
before

> After
after

5. The reason for the tuple of exception class in that parameter. I edited the `` to adjust for now.
import

> I'm not so sure with `from requests.packages.urllib3.exceptions import ProtocolError` they indirectly suggested in #1606 since it has the same exception class as `from urllib3.exceptions import ProtocolError` although Gept said it was different. Therefore it might not cover the problem ideally. I might suggest using the base exception class of all modules.
diff

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the
+  overall community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or
+  advances of any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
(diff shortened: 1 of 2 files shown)


Harmon758 (commented):
First of all, please don't make pull requests with multiple separate and independent changes. The code of conduct file should not be part of this PR.

Secondly, please do not use chatbots to contribute to codebases that you're not familiar with. This results in low-effort and low-quality PRs that are not useful and waste maintainer time and effort to handle.

This PR specifically does not resolve #1606, and in fact, doesn't touch anything related to `API`. It also introduces an unnecessary dependency and doesn't add it as a requirement, which would result in errors on import. Regardless, this is not the way to handle/fix this issue. I've responded to your support post in Discord, with methods to handle your issue.

ranggakd:
Hi @Harmon758,

Thanks for your feedback on my pull request. I appreciate the constructive criticism.

However, I think it would be helpful to have a pull request template for contributors who are not familiar with the codebase. This could help to avoid low-quality PRs and reduce the time and effort required by maintainers to handle them.

Perhaps this is something that could be added to the repository's documentation or issue templates. In any case, I will keep this in mind for future contributions.

PR #2087, opened 2023-03-27
Create Funny Quotes

KAIAHRi:


Harmon758:
This is not something that belongs in the library, especially not at the root level.
If you're testing out pull requests, please use your own repository to do so.

Issue #2096, opened 2023-04-25
Display time remaining after you hit the rate limit


thecoderenroute:
When pulling a lot of data, it becomes a tedious task to constantly hit the rate limit and then not know when it will resume. A mechanism through which we could see the number of seconds remaining or a progress bar or a visualization of any kind to indicate the remaining wait time would be extremely useful.

RFlorindo:
Can I take this?

PR #2121, opened 2023-05-27
# Add progress bar to track the sleep time after "Rate limit reached" warning

RFlorindo:
This change resolves issue #2096 by adding a progress bar to track the sleep time after reaching the rate limit of Twitter API.

### DIFF
--- a/tweepy/client.py
+++ b/tweepy/client.py
@@ -4,6 +4,8 @@
 
 from collections import namedtuple
 import datetime
+import time
+from tqdm import tqdm
 
 try:
     from functools import cache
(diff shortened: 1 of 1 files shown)


Harmon758:
This adds and uses an unnecessary external dependency without actually adding it as a requirement for the library. This would break any imports of the library if installed in an environment without tqdm already installed.

Additionally, it adds an undocumented output to `stderr` that isn't consistent with the rest of the library. This is not how #2096 should be resolved.

Even if the above were not true, this implementation ends up sleeping twice as long as is necessary to handle the rate limit.

I'm also not sure why there's a redundant `time` import being added or an unrelated f-string being reduced, and the import grouping being added is inconsistent here and with the rest of the library.

RFlorindo:
Thank you for your comments Hammon758, you are completely right. I am newcomer to this, I will take your observations into consideration, do the necessary adjustments, and try again later if that is alright with you.

I'm not sure how to test this yet, but I'll  find a way. I'll proceed with the pull request only after ensuring that all the necessary conditions are met.

Once again sorry and thank you.

PR #2134, opened 2023-06-20
Edit redme

spandhare11:


Harmon758:
Tweepy is a Python library used to interface with Twitter's API, not a Twitter clone.

PR #2173, opened 2024-01-30
Fix minor documentation inconsistency in authentication.rst

horatiothomas:
Reference variable rather than string as is done for other parameter


horatiothomas:
Actually shouldn't this documentation be updated to reference the new initializer?

Issue #2177, opened 2024-03-19
imghdr deprecation warning in Python 3.11


primal100:
### Summary

imghdr is deprecated for removal in python 3.13

### Reproduction Steps

Import the library in python 3.11

### Minimal Reproducible Example

```python
venv311\Lib\site-packages\tweepy\api.py:7
  <directory>\venv311\Lib\site-packages\tweepy\api.py:7: DeprecationWarning: 'imghdr' is deprecated and slated for removal in Python 3.13
    import imghdr
```

### Expected Results

No deprecation warning

### Actual Results

Deprecation warning

### Twitter API Access Plan

Free

### Tweepy Version

Version: 4.14.0

### Checklist

- [X] I have searched for duplicate issues.
- [X] If applicable, I have shown the entire traceback.
- [X] If applicable, I have removed any visible credentials from my code and/or screenshots.

### Additional Context

Suggestion would be to use only mimetypes to detect file type (currently both imghdr and mimetypes are used)

cclauss:
* #2181

endlisnis:
No commits in 10 months and 77 open bugs.  Sounds like this project is dying.  Too bad, I use this regularly.

joshthecoder:
Fixed in v4.15.0

PR #2181, opened 2024-05-11
GitHub Actions: Test on Python 3.13 beta

cclauss:
Confirms:
* #2177

Related:
* #1086 -- ~@shuuji3 Would you be able to propose a new fix?~

https://www.python.org/downloads/release/python-3130b1/

Raises `ModuleNotFoundError: No module named 'imghdr'` because Python 3.13 removes it from the Standard Library.
* https://docs.python.org/3/library/imghdr.html

---

> imghdr: use the projects [filetype](https://pypi.org/project/filetype/), [puremagic](https://pypi.org/project/puremagic/), or [python-magic](https://pypi.org/project/python-magic/) instead. (Contributed by Victor Stinner in [gh-104773](https://github.com/python/cpython/issues/104773).)

https://docs.python.org/3.13/whatsnew/3.13.html#pep-594-dead-batteries-and-other-module-removals

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,14 +7,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
 
     steps:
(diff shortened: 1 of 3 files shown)


radarhere:
I've created https://github.com/tweepy/tweepy/pull/2205 as an alternative, starting with this commit, but falling back to `mimetypes` instead of raising an error.

PR #2205, opened 2024-10-06
Fallback to mimetypes if imghdr is not available

radarhere:
https://github.com/tweepy/tweepy/pull/2181 started the ball rolling by suggesting testing Python 3.13, and raising a `ModuleNotFoundError` if `imghdr` could not be imported when it was to be used.

Since the code already falls back to `mimetypes` if `imghdr` did not return anything
https://github.com/tweepy/tweepy/blob/91a41c6e1c955d278c370d51d5cf43b05f7cd979/tweepy/api.py#L3476-L3480

this PR suggests also falling back to `mimetypes` if `imghdr` is not available at all. The test suite still passes like this.

It does not handle filenames without an extension, like #1086 did, but I imagine it's better than always raising an error in Python 3.13.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,14 +7,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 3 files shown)


joshthecoder:
@radarhere @cclauss thanks for the contribution here 👏

PR #2208, opened 2024-11-05
Update api.py

Neecrownsmith:
Update imghdr with Image(pillow) as imghdr is deprecated in current python version (3.13.0)


joshthecoder:
Fixed in 4.15.0 release

